PR #14, opened 2024-04-22
Feature/gsye 717 gsy node

spyrostz:
Please review only the gsy-node/ and primitives/ changes for now. Matching engine still needs updates


gsy-matching-engine/src/connectors/redis_connector.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use anyhow::{Error, Result};
+use serde_json::{Value, json};
+use chrono::{NaiveDateTime};
+use redis::Commands;
+
+pub fn value_to_str(value: &Value) -> String {
+    // Helper function to convert the serde Value to String
+    match value.as_str() {
+        Some(..) => value.as_str().unwrap().to_string(),
```
hannesdiedrich, line 12: I `unwrap()` needed here, because I understand that you handle the None case already in `match`?

hannesdiedrich: ah no, please explain what is happening here. 
I understand the following steps:
- `as_str()`: `value` is converted into a str
- 'unwrap()': if the outcome is None, return a panic (which I understand is an Exception in python)
- 'to_string()': convert the outcome (be it a panic or a string) to a string in order to be compliant with the output type of the function
Did I get it right?

spyrostz: The latter message is right, to explain a bit more details:
- `match` operator performs pattern matching. This splits the flow according to the type of `value.as_str()`. If the type of the return value is None, then it creates a new String(). `Some` is a type that denotes that a value has returned, but we need to unwrap it in order to retrieve it. Admittedly, I think that this:
```
    match value.as_str() {
        Some(str_val) => str_val.to_string(),
        None => String::new(),
    }
``` 
would achieve exactly the same, and I will fix it on the matching-engine PR. 
- `str` and `String` are different types in Rust. In general, str is more efficient when passing through functions, since it is an immutable reference to a list of characters. If the list of characters needs to be modified, the String needs to be used, which is mutable and allows you to modify it, since it is not a reference but an object that owns the data that it handles. Wanted to clarify this because this is not just to satisfy the signature of the function, but in order to return a "string" that can be modified at will.

gsy-matching-engine/src/connectors/substrate_connector.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let orderbook_url = Arc::new(Mutex::new(orderbook_url));
+    let node_url = Arc::new(Mutex::new(node_url.clone()));
+
+    while let Some(Ok(block)) = gsy_blocks_events.next().await {
+        info!("Block {:?} finalized: {:?}", block.number, block.hash());
+
+        let matches = Arc::new(Mutex::new(Vec::new()));
+
+        if (block.number as u64) % 4 == 0 {
```
hannesdiedrich, line 48: Can you please explain why you do a mod 4 here? Thanks!

gsy-matching-engine/src/algorithms/mod.rs
hannesdiedrich, line 1: Can you please explain why this `algorithms` folder is needed?
AFAIS, only one trait is in it.

gsy-matching-engine/src/connectors/substrate_connector.rs
```diff
@@ -0,0 +1,193 @@
...
+    let order_submit_and_watch = order_transfer
+        .sign_and_submit_then_watch_default(&signer)
+        .await?
+        .wait_for_finalized_success()
+        .await?;
+
+    let transfer_event =
+        order_submit_and_watch.find_first::<gsy_node::trades_settlement::events::TradeCleared>()?;
+
+    if let Some(event) = transfer_event {
```
hannesdiedrich, line 158: I am really having trouble understanding this. I found https://stackoverflow.com/questions/72081657/rust-lang-what-is-if-let-somex-x-doing but i can not translate it to this case here.
I do not see where `event` is assigned or declared. Can you please help me?

hannesdiedrich (approved):
I had a broader look into this and tried to understand the general structure. I think I have an idea now. Still, I left a few questions. Even though I know that you just moved the code here, would be good to discuss. Thanks a lot! Already approving

gsy-matching-engine/src/connectors/substrate_connector.rs
spyrostz: This is trying to configure the matching frequency actually. Essentially it says that it will perform the matching operation every 4 blocks. I will rename it in order to be more easily readable

gsy-matching-engine/src/algorithms/mod.rs
spyrostz: It is there in order to accomodate different matching algorithms. I would leave it even if now it is a very small package.

gsy-matching-engine/src/connectors/substrate_connector.rs
spyrostz: This is how `if let` works in Rust https://doc.rust-lang.org/book/ch06-03-if-let.html. Essentially it declares the `event` variable that is populated if the `transfer_event` is `Some` value and not an error.

gsy-matching-engine/src/connectors/substrate_connector.rs
```diff
@@ -40,7 +42,7 @@ pub async fn substrate_subscribe(orderbook_url: String, node_url: String) -> Res
 
         let matches = Arc::new(Mutex::new(Vec::new()));
 
-        if (block.number() as u64) % 4 == 0 {
+        if (block.number() as u64) % MATCH_PER_NR_BLOCKS == 0 {
```
hannesdiedrich, line 45: ❤️

BigTava: ❤️

PR #15, opened 2024-04-29
Feature/gsye 717 orderbook service

spyrostz:


gsy-node/modules/orderbook-worker/src/benchmarking.rs
```diff
@@ -13,106 +13,106 @@ use frame_system::RawOrigin;
 use frame_support::{sp_runtime::traits::{Hash, One}};
 use sp_std::vec;
 
-fn add_user<T: Config>(
```
hannesdiedrich: So you just commented it out in order to not to having to adapt it because it is "just" benchmarking?

gsy-orderbook-service/src/event_listener.rs
```diff
@@ -1,73 +1,79 @@
...
-            match db
-                .get_ref()
-                .orders()
-                .update_order_status_by_id(id, OrderStatus::Executed)
-                .await
-            {
-                Ok(result) => info!("Update result: {:?}", result),
-                Err(e) => {
-                    tracing::error!("Failed to execute update: {:?}", e);
+                let id = &bson::to_bson(&order_executed.0.offer_hash).unwrap();
```
hannesdiedrich, line 35: Do I get it right, that this is the place where the trade is written to mongo?

hannesdiedrich (approved):
LGTM

gsy-orderbook-service/src/event_listener.rs
spyrostz: Exactly, this changes the status of the order from Open to Executed.

gsy-node/modules/orderbook-worker/src/benchmarking.rs
spyrostz: Exactly, I plan to reintroduce it in a separate PR now that the dependencies are in good shape

PR #18, opened 2024-08-12
DD-349: Integrated the matching engine to the docker compose file. Ad…

spyrostz:
…apted the matching engine to use newer Rust and respective packages (e.g. SubXT latest version).


BigTava (approved):
Really nice work! You even updated rust to 1.80

PR #39, opened 2025-12-22
INTELLIGENT: Implement degrees-of-freedom in two sided pay-as-clear and pay-as-bid algorithms

keyanhov:


e2e-tests/src/steps/common_steps.rs
```diff
@@ -28,6 +28,7 @@ async fn services_are_running(world: &mut MyWorld) {
 #[given(
 	regex = r#"users "([^"]*)", "([^"]*)", and "([^"]*)" the matching engine operator are registered and have collateral"#
 )]
+#[given(regex = r#"users "([^"]*)", "([^"]*)", and "([^"]*)" are registered and have collateral"#)]
 async fn users_are_registered(
 	world: &mut MyWorld,
 	alice_name: String,
```
spyrostz: I think that it makes sense to rename the 3 variables to `first_user`, `second_user` and `third_user` respectively.

e2e-tests/src/steps/market_steps.rs
```diff
@@ -1,100 +1,110 @@
...
-use std::time::Duration as Duration;
-use subxt::utils::H256;
+use gsy_offchain_primitives::utils::string_to_h256;
+use gsy_offchain_primitives::{constants::GLOBAL_CONSTANTS, MarketType};
+use std::time::Duration;
 use tokio::time::sleep;
-use tracing::{error, info};
-
-#[when(regex = r#"the community topology and forecasts of (\d+) energy are submitted"#)]
-async fn submit_topology_forecasts(world: &mut MyWorld, energy: f64) {
```
spyrostz, line 16: I think that we will have some merge conflicts here

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -1,52 +1,51 @@
...
-use gsy_offchain_primitives::db_api_schema::{
-    profiles::MeasurementSchema, 
-    trades::TradeSchema,
-};
-use gsy_offchain_primitives::constants::GlobalConstants;
+use tracing::info;
 
 fn round_down_timeslot(ts: u64) -> u64 {
-    (ts / GlobalConstants.TIME_SLOT_SEC) * GlobalConstants.TIME_SLOT_SEC
+	(ts / GLOBAL_CONSTANTS.time_slot_sec) * GLOBAL_CONSTANTS.time_slot_sec
```
spyrostz, line 8: Why rename these? If it is for the warnings, these were deliberately disabled in the definition of the GlobalConstants struct:

`#![allow(non_snake_case,non_upper_case_globals)]`

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -1,52 +1,51 @@
...
 }
 
 pub async fn fetch_trades_and_measurements_for_timeslot(
-    base_url: &str,
-    timeslot: u64,
-    market_duration: u64,
+	base_url: &str,
```
spyrostz, line 12: Are these tabs / spaces diffs an artifact of the IDE ? If yes, I would run `cargo fmt` at the end of each diff in order to harmonise them.

offchain-primitives/src/types.rs
```diff
@@ -60,45 +83,153 @@ impl PayAsBid for MatchingData {
 	type Output = BidOfferMatch;
 
 	fn pay_as_bid(&mut self) -> Vec<Self::Output> {
-		let mut bid_offer_pairs = Vec::new();
+		let mut matches = Vec::new();
+		let available_bids = self.bids.clone();
+		let available_offers = self.offers.clone();
+
+		// --- Phase 1: Preference-based Matching ---
```
spyrostz: I would leave the pay_as_bid method as is, in order to not introduce regressions and to be able to use it in isolation. A separate method could be created that would only perform the attribute matching, and its results could be filtered out of the input to `pay_as_bid` method.

offchain-primitives/src/types.rs
```diff
@@ -60,45 +83,153 @@ impl PayAsBid for MatchingData {
...
+		let mut preference_matches = Vec::new();
+		let mut remaining_bids_after_pref = Vec::new();
+		// Use H256 (hash of the order) to track matched amounts
+		let mut bid_matched_amounts: HashMap<H256, u64> = HashMap::new();
```
spyrostz: Since the nonce is a unique identifier, why not using it for now ? Ideally we should replace the nonce with something more inherently-unique (e.g. a uuid) but for now it should work. 

Reason I am suggesting this, is because if the order is mutated then it will lead to a different hash value, which will then result in not being able to fetch it from these dicts.

spyrostz (commented):
Left some comments, will review again after the formatting comment is addressed.

gsy-execution-engine/src/connectors/offchain_storage.rs
keyanhov: I decided to proceed with the renaming to align with standard Rust naming conventions (RFC 430) rather than relying on compiler suppressions (#[allow(...)]).

PR #40, opened 2026-01-05
Feature/dd 367

spyrostz:


offchain-primitives/src/constants.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(non_snake_case,non_upper_case_globals)]
```
hannesdiedrich, line 1: This is for the linter of the compiler and is equivalent to the `pylint: disabler= ...`, correct?

gsy-community-client/src/node_connector/orders.rs
```diff
@@ -17,113 +17,118 @@ const OFFER_RATE: f64 = 0.07;
...
-	let input_orders = create_input_orders(forecasts, market, signer);
-	let register_order_tx = gsy_node::tx().orderbook_worker().insert_orders(input_orders);
+    let input_orders = create_input_orders(forecasts, market, signer);
+    let register_order_tx = gsy_node::tx()
+        .orderbook_worker()
+        .insert_orders(input_orders);
 
-	let order_submit_and_watch = api
-		.tx()
-		.sign_and_submit_then_watch_default(&register_order_tx, signer)
```
hannesdiedrich, line 32: Can you please explain the reason for these whitespace changes?

hannesdiedrich: I see now. This is done by `cargo fmt`, right?

gsy-community-client/src/external_api.rs
```diff
@@ -22,18 +22,92 @@ pub struct ExternalMeasurement {
...
+    pub asset_sub_type: Option<NameField>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct LECCommunityAssetResults {
+    pub bindings: Vec<ExternalCommunityAsset>,
+}
+
+#[derive(Deserialize, Debug, Clone)]
+pub struct GetLECAssets {
```
hannesdiedrich: I was confused by the name at first and thought that this is somehow getting the LEXAssets. AFAIU, this is "only" the type of the results of the assets API. Right?
If yes, I would rename to `LECAssetsResults` or similar. What do you think?
The same applies to `GetLECBuildings`.

gsy-community-client/src/external_api.rs
hannesdiedrich, line 1: This is a NIT, but I was also confused by the name of the module. 
This is really only a collection of utils and types, right? The name suggests that these are external api implementations.

gsy-community-client/src/main.rs
```diff
@@ -1,151 +1,235 @@
...
+    }
+
+    // Function to fetch an array of measurement data
+    async fn fetch_measurements(&self) -> Result<Vec<ExternalMeasurement>, reqwest::Error> {
+        let response = self.client.get(&self.measurements_url).send().await?;
+        response.json::<Vec<ExternalMeasurement>>().await
+    }
+
+    async fn fetch_topology(&self) -> Result<GetLECBuildings, reqwest::Error> {
+        let response = self.client.get(&self.topology_url).send().await?;
```
hannesdiedrich: As discussed on slack
```suggestion
        let response = self.client.post(&self.topology_url).send().await?;
```

gsy-community-client/src/external_api.rs
```diff
@@ -22,18 +22,92 @@ pub struct ExternalMeasurement {
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct NameField {
+    #[serde(rename = "type")]
+    pub field_type: String,
+    pub value: String,
+}
+
+// Struct for forecast data received from external API
```
hannesdiedrich, line 71: Please remove or adapt

gsy-community-client/src/external_api.rs
```diff
@@ -22,18 +22,92 @@ pub struct ExternalMeasurement {
...
 // Struct for forecast data received from external API
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, Hash, Eq, PartialEq)]
 pub struct ExternalAreaTopology {
-    pub area_uuid: String,
     pub area_name: String,
+    pub area_type: AssetType,
 }
 
 // Struct for forecast data received from external API
```
hannesdiedrich, line 57: Please remove or adapt

hannesdiedrich (approved):
I focused on the community client side and left some minor comments.
I could follow the code there and say LGTM.

offchain-primitives/src/constants.rs
spyrostz: Exactly. This disables the warnings for: 
- variables not being snake-case only (therefore it allows variables such as active_power_kW which has an uppercase character)
- allows static variables to not be uppercase-only, thus allowing naming variables such as `GlobalConstants`

gsy-community-client/src/node_connector/orders.rs
spyrostz: Exactly. I generally run `cargo fmt` at each commit, in order to harmonize the format of the source code.

gsy-community-client/src/external_api.rs
spyrostz: Agreed and correcting both struct names, thanks!

gsy-community-client/src/main.rs
spyrostz: Fixing, thanks!

gsy-community-client/src/external_api.rs
spyrostz: This is indeed the forecast struct that we expect from InfluxDB API, so IMO the comment still stands. If you think it is an overkill to explain it here then we can remove it, let me know to do so if you agree

gsy-community-client/src/external_api.rs
spyrostz: This is a collection of the data types that we receive from the external APIs, that's why I named the module like this. I could rename it to `external_api_types.rs` if you also think it would be clearer.

hannesdiedrich: Ah no. From todays point of view the existing name is OK. Thanks!

gsy-community-client/src/external_api.rs
hannesdiedrich: I understand that this struct is representing the houses in a community. 
This topology is used in the forecast acquisition, yes, but is not only used there. Or am I missing something?

e2e-tests/src/steps/community_client_steps.rs
```diff
@@ -0,0 +1,27 @@
+use crate::world::MyWorld;
```
hannesdiedrich, line 1: I am having trouble finding sufficient documentation for this crate. Consequently, I have trouble understanding the role of it here in the step implementations. Is `MyWorld` the type of the test context?

hannesdiedrich (approved):
LGTM

e2e-tests/src/steps/community_client_steps.rs
spyrostz: Exactly, the `MyWorld` struct is similar to `context` in Python behave. We use it to add attributes that should be shared across test steps.

PR #44, opened 2026-03-05
Matching service

spyrostz:


launchpad-matching-service/openapi.yaml
hannesdiedrich, line 1: Did AI add this automatically?

hannesdiedrich: What I meant is: Did you use AI to create this file?

spyrostz: Yes, I used it to autogenerate the OpenAPI documentation

launchpad-matching-service/src/api/controller.rs
```diff
@@ -0,0 +1,40 @@
...
+            Order::Bid(bid) => bid.bid_component.market_id.clone(),
+            Order::Offer(offer) => offer.offer_component.market_id.clone(),
+        }
+    }).collect();
+
+    for market_id in market_ids.iter() {
+        let bids_list: Vec<DbBid> = orders.iter().filter_map(|order| {
+            match &order.order {
+                Order::Bid(bid) if bid.bid_component.market_id == *market_id => Some(bid.clone()),
+                _ => None,
```
hannesdiedrich: This case should never be reached, right? Because the market Id was just obtained by the list of orders beforehand right? Or am I missing something?

launchpad-matching-service/src/api/types.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        let mut bids = self.bids.clone();
+        let mut offers = self.offers.clone();
+
+        bids.sort_by(|a, b| {
+            b.bid_component
+                .energy_rate
+                .partial_cmp(&a.bid_component.energy_rate)
+                .unwrap()
+        });
```
hannesdiedrich, line 56: Very interesting. I understand that this sorts the bids in descending order WRT the energy rate. AI helped me understand, but do you agree that this is not really readable for a human? Do you mind adding an inline comment , also for the sorting of the offers?
Thanks!

launchpad-matching-service/src/api/types.rs
```diff
@@ -0,0 +1,143 @@
...
+use gsy_offchain_primitives::db_api_schema::orders::{
+    DbBid, DbOffer, DbOrderComponent, DbOrderSchema,
+};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use std::collections::HashMap;
+use subxt::config::{Hasher, substrate::BlakeTwo256};
+use subxt::utils::H256;
+
+const EPS: f64 = 0.000001;
```
hannesdiedrich, line 11: I understand that this is equivalent to our FLOATING_POINT_TOLLERANCE in python. Is EPS short for epsilon or is this a usual rust naming for such a concept?

launchpad-matching-service/src/api/controller.rs
```diff
@@ -0,0 +1,388 @@
...
+                .collect();
+            let mut matching_data = DbMatchingData {
+                bids: bids_list,
+                offers: offers_list,
+                market_id: market_id.clone(),
+                user_id: user_id.clone(),
+            };
+            let algorithm_result = matching_data.pay_as_bid();
+            all_matches_to_insert.extend(algorithm_result.clone());
+            matches.insert(market_id.clone(), algorithm_result);
```
hannesdiedrich: Minor:
Why do we need both `all_matches_to_insert` and `matches`? isn't there such thing like the equivalent of `list(keys(...))` in python?

launchpad-matching-service/src/api/controller.rs
```diff
@@ -0,0 +1,388 @@
...
+                .unwrap_or(0.0);
+
+            entry.total_matched_energy_kWh = matched_energy;
+            entry.total_unmatched_energy_kWh = bid_energy + offer_energy - 2.0 * matched_energy;
+        }
+
+        market_data_map
+    }
+
+    async fn insert_bid_offer_matches_to_db(&self, matches: Vec<DbBidOfferMatch>);
```
hannesdiedrich, line 182: I understand that this still needs to be implemented?

hannesdiedrich: All these async functions right?

hannesdiedrich: Ignore me, this is the "Baseclass" and they are implemented in the child class. 
(Sorry for keep using python numenclature :-) )

launchpad-matching-service/src/api/controller.rs
```diff
@@ -0,0 +1,388 @@
...
+    async fn get_market_statistics(
+        &self,
+        user_id: String,
+        market_id: Option<String>,
+        start_time: u64,
+        end_time: u64,
+        resolution: model::Resolution,
+    ) -> model::MarketStatisticsResponse {
+        let mut response = model::MarketStatisticsResponse {
+            average_trade_rate_timeseries: Vec::new(),
```
hannesdiedrich, line 268: AFAIU. this could also he a HashMap, couldn't it?
Is there a best practice reason do have a list of a specific type for this instead of a hashmap?

hannesdiedrich (approved):
I have reviewed as best as I could and could at least follow the majority of code. 
I left some questions, but from my point of view this is good to go. Thanks a lot!

launchpad-matching-service/src/api/types.rs
spyrostz: It is short for epsilon indeed, which is the standard math term to define tolerance. I wanted something less verbose than FLOATING_POINT_TOLERANCE here (which I do not like that it is that verbose in gsy-e)

launchpad-matching-service/src/api/controller.rs
spyrostz: Main reason is that the return value makes more sense to be grouped per market for the response of this endpoint, while for the database it is more appropriate to save flat data and not group them by market_id. Since this endpoint produces matches from different markets, it makes sense to return the values grouped per market (this is the `matches` variable). The database though has no concept of mapping - essentially we need to save a vector of matches and find the mapping of these with market_id via database queries (this is the `all_matches_to_insert`)

launchpad-matching-service/src/api/controller.rs
spyrostz: Exactly, this is similar to an abstract method in Python. FYI this concept is called `trait` in Rust

launchpad-matching-service/src/api/controller.rs
spyrostz: Usually it is more convenient for FE to not receive a datetime -> object HashMap, but a Vec of objects when parsing timeseries. Main reason is that most graph libraries consume JSON arrays and not JSON objects / mappings. Therefore I thought to do their life easier and return a Vec of objects.

launchpad-matching-service/src/api/types.rs
spyrostz: I will leave a comment indeed, this is pretty standard sorting operation for Rust but does not hurt to add.

launchpad-matching-service/src/api/controller.rs
spyrostz: This should indeed never be reached, but the `match` branch obliges us to handle all cases in an Enum, otherwise compilation fails.

launchpad-matching-service/src/api/controller.rs
hannesdiedrich: I see, thanks a lot for (all) explanation(s)!

launchpad-matching-service/src/api/controller.rs
```diff
@@ -0,0 +1,388 @@
...
+        &self,
+        orders_obj: OrdersToMatch,
+    ) -> HashMap<String, Vec<DbBidOfferMatch>> {
+        let orders = orders_obj.orders;
+        let user_id = orders_obj.user_id;
+        let mut matches = HashMap::new();
+        let mut all_matches_to_insert = Vec::new();
+
+        // Find all market ids in the orders
+        let market_ids: Vec<String> = orders
```
keyanhov: Should we deduplicate `market_ids` before the loop?
Right now it’s built with one entry per order, so the same market_id can appear multiple times, which means we re-process the same market repeatedly and append duplicate entries to all_matches_to_insert (inflating persisted matches/statistics). Would using a unique set of market IDs (HashSet or sort_unstable + dedup) make sense here?

launchpad-matching-service/src/api/types.rs
```diff
@@ -0,0 +1,147 @@
...
+                .energy_rate
+                .partial_cmp(&b.offer_component.energy_rate)
+                .unwrap()
+        });
+
+        let mut available_order_energy: HashMap<H256, f64> = HashMap::new();
+
+        for offer in &mut offers {
+            for bid in &mut bids {
+                if offer.offer_component.area_uuid == bid.bid_component.area_uuid
```
keyanhov, line 70: Here we're not enforcing that bid and offer belong to the same market slot window (i.e. 15 min), only checking market/area/price? Should we compare normalized buckets (e.g. time_slot / 900) to avoid matching orders from different market slots in the same market

docker-compose.launchpad-matching-service.yml
```diff
@@ -0,0 +1,48 @@
...
+    build:
+      context: .
+      dockerfile: ./launchpad-matching-service/Dockerfile.matching-service
+    depends_on:
+      - mongodb
+    environment:
+      - RUST_LOG=info
+      - DATABASE_HOST=mongodb
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/health-check"]
```
keyanhov: in `Dockerfile.matching-service` the service binds to exposed port 9876

launchpad-matching-service/src/api/model.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub client: Client,
+    pub db: mongodb::Database,
+    pub collection_name: String,
+}
+
+impl MatchModel {
+    pub async fn new() -> mongodb::error::Result<Self> {
+        let mongodb_uri = get_configuration().unwrap().get_connection_string();
+        let client = Client::with_uri_str(mongodb_uri).await?;
+        let db = client.database("launchpad");
```
keyanhov: we can use database_name from the config.

```
let configuration = get_configuration().expect("Failed to load configuration");
let db = client.database(configuration.database_name);
```

launchpad-matching-service/src/auth/model.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub struct UserModel {
+    db: mongodb::Database,
+}
+
+impl UserModel {
+    pub async fn new() -> mongodb::error::Result<Self> {
+        let mongodb_uri = get_configuration().unwrap().get_connection_string();
+        let client = Client::with_uri_str(mongodb_uri).await?;
+        let db = client.database("launchpad");
```
keyanhov: we can use database_name from the config.

```
let configuration = get_configuration().expect("Failed to load configuration");
let db = client.database(configuration.database_name);
```

launchpad-matching-service/tests/views.rs
```diff
@@ -0,0 +1,725 @@
...
+    let app = test::init_service(App::new().service(views::health_check)).await;
+
+    let req = test::TestRequest::get().uri("/health-check").to_request();
+    let resp = test::call_service(&app, req).await;
+
+    assert!(resp.status().is_success());
+}
+
+#[actix_web::test]
+async fn test_pay_as_bid_endpoint() {
```
keyanhov, line 81: Should we update these endpoint tests to include JwtSecret in app_data and send an Authorization: Bearer ... header, so they exercise the real protected-route behavior with Claims extraction?

keyanhov (commented):
Added few comment once addressed LGTM

launchpad-matching-service/src/api/controller.rs
spyrostz: Good catch, a HashSet would be correct here indeed

launchpad-matching-service/src/api/types.rs
spyrostz: In our initial design we assumed that there is only one time slot per market id, therefore there was no need. This service indeed though accepts bids and offers directly for the user, therefore it makes sense to distinguish both. Fixed

docker-compose.launchpad-matching-service.yml
spyrostz: You are correct, fixing

launchpad-matching-service/src/api/model.rs
spyrostz: Fixing

launchpad-matching-service/src/auth/model.rs
spyrostz: Same as above, corrected here too

launchpad-matching-service/tests/views.rs
spyrostz: You are correct, fixed the tests to include JWT authentication

PR #46, opened 2026-06-03
EWDS Integration

keyanhov:


e2e-tests/src/steps/market_steps.rs
```diff
@@ -1,135 +1,137 @@
...
+
+abigen!(
+    MarketControllerContract,
+    r#"[
+        function isMarketOpen(bytes32 marketId) external view returns (bool)
+    ]"#
+);
+
+#[when(
+    regex = r#"^the community topology and forecasts of (\d+) energy are submitted by "([^"]*)", "([^"]*)", and "([^"]*)"$"#
```
hannesdiedrich: AFAIS, regex is not required for the pattern matching here. Do you mind if we use human readable ( ;-) ) pattern matching?
I think the following would also work:
```
#[when("the community topology and forecasts of {energy} energy are submitted by {user1}, {user2}, and {user3}")]
```

docker-compose.ewds.yml
```diff
@@ -0,0 +1,217 @@
...
+    image: nginx:alpine
+    volumes:
+      - ./proxy-config:/etc/nginx/conf.d:ro
+    ports:
+      - "${EWDS_GATEWAY_PROXY_PORT:-3009}:80"
+    depends_on:
+      - ddhub-gateway-UI
+      - ddhub-gateway-api
+
+  gsy-offchain-storage:
```
hannesdiedrich: Why is there no mongo db connection?
I also do not see a mongo-db pod in here. 
Or am I missing something?

e2e-tests/src/steps/market_steps.rs
keyanhov: Agreed that a Cucumber Expression is more readable than the regex. The exact snippet needs a small adjustment though: in cucumber-rs the bare string form is an exact literal, so it won’t capture values. We should use `expr = ...`, e.g. `{float}` for energy and `{string}` for the quoted user names.
i.ei
```
#[when(
    expr = "the community topology and forecasts of {float} energy are submitted by {string}, {string}, and {string}"
)]
```

docker-compose.ewds.yml
keyanhov: the compose actually was mostly used to integrate the ewds env config and to run together with the `docker-compose` and `docker-compose.test`. We could remove GSY service overrides from `docker-compose.ewds.yml` and move ewds mode GSY env configuration into the normal compose files behind env-var defaults. wdyt?

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -0,0 +1,672 @@
...
+        config,
+        config.response_fqcn.clone(),
+        request_id,
+        topic_name,
+        payload,
+    )
+    .await
+}
+
+async fn send_message_with_fqcn(
```
hannesdiedrich, line 605: I never heard about but researched for FQCN and understand that you relate to the Fully Clasified Class Name in PHP, right?
I searched for a equivalent in Rust, but I understand now, that you mean a absolute path or url, right?

hannesdiedrich: Forget me, I see this is a nomenclature that is used in energy web.

e2e-tests/src/steps/market_steps.rs
hannesdiedrich: Ah yes, you are right! Thanks a loot for agreeing!

docker-compose.ewds.yml
hannesdiedrich: I am not quite sure what you suggest. For the staging environment, we need the CGW and the offchain storage deployed, for this I am creating a new dockerfile where I copy the EWDS services and configuration from docker-compose.ewds.yml and add the offchain storage and mongodb pods in there.
For the offchain storage I add the configuration parameters from the tests compose file AND from the ewds compose file. I attached my proposal. What do you think?


[docker-compose.staging.yml](https://github.com/user-attachments/files/28709872/docker-compose.staging.yml)

hannesdiedrich: Sorry, the dependency to gsy-contracts-bootstrap and anvil should be removed.
Here is my proposal that is currently running on my machine.
https://github.com/gridsingularity/gsy-decentralized-exchange/commit/90bdbb789ee60f4970d09351f5326461d7ef2e02

keyanhov: I think creating a separate docker-compose.staging.yml it's a good idea. I've already pushed few changes to modify both the cucumber regex and the compose in order to remove from docker-compose.ewds.yml the gsy dex services.

hannesdiedrich: So you propose to trigger two docker compose?
```
docker compose --env-file .env.ewds.local -f docker-compose.ewds.yml up
docker compose -f docker-compose.yml up
```
In order to start all services?

keyanhov: correct. It's important to ensure that the EWDS related services start (without errors) and are correctly configured before going ahead with starting our services

hannesdiedrich: I see. You mean that the manual adding of MTLS certificate and the channels and topics needs another restart of the EWDS compose. Only once this is configured, we can start the other services, right?

docs/setup/docker.md
hannesdiedrich, line 1: I would rename this file to ewds or similar. What do you think?

gsy-offchain-storage/src/evm_handler.rs
```diff
@@ -0,0 +1,147 @@
...
+use gsy_offchain_primitives::utils::{bytes16_to_hex, NODE_FLOAT_SCALING_FACTOR};
+use tracing::{error, info, warn};
+
+pub struct OffchainStorageEvmHandler {
+    pub db: DatabaseWrapper,
+}
+
+#[async_trait]
+impl GsyEventHandler for OffchainStorageEvmHandler {
+    async fn handle_order_placed(&self, event: OrderPlacedFilter) -> Result<()> {
```
hannesdiedrich, line 21: Sorry to ask something that fundamental, but are orders really places on the block chain?
I thought that these are send via the EWDS and received by all listening services.

hannesdiedrich: Ah I see the if block in [main.rs](https://github.com/gridsingularity/gsy-decentralized-exchange/blob/ad7fc4b1ab50e03a1495cd0fadd80d7b420d03b5/gsy-offchain-storage/src/main.rs) now. If EWDS_ENABLE_HANDLER is true, then we communicate via EWDS. So I understand that you wanted to keep the EVM subscription, too. But now there are two processes that are reacting to a creation of an order. Did I get it right? Is this what we want?

gsy-offchain-storage/src/startup.rs
hannesdiedrich, line 22: Again a general question:
I understand that this http server that exposes the endpoints listed below is started here. 
but where do we call the function "run"? I would have expected inside the main.rs but there I do not see it.

gsy-offchain-storage/src/main.rs
```diff
@@ -18,27 +21,55 @@ async fn main() -> Result<(), anyhow::Error> {
...
         "{}:{}",
         configuration.application_host, configuration.application_port
     );
     let listener = TcpListener::bind(address).expect("Failed to bind");
+
+    info!("Server listening on {}", listener.local_addr().unwrap());
+
     match run(listener, db_connection_wrapper)?.await {
         Ok(_) => Ok(()),
         Err(e) => Err(Error::from(e)),
```
hannesdiedrich, line 75: I know this is not part of the PR, but maybe Spyros: What do we need this TCP listener for here?

gsy-offchain-storage/src/main.rs
```diff
@@ -18,27 +21,55 @@ async fn main() -> Result<(), anyhow::Error> {
...
+            db: db_for_listener,
+        };
+        let listener = GsyEthersListener::new(listener_config, handler);
+        if let Err(e) = listener.run().await {
+            tracing::error!("EVM Listener crashed: {:?}", e);
+        }
+    });
+
+    tokio::task::spawn(async move {
+        start_scheduler(db, scheduler_interval).await;
```
hannesdiedrich, line 53: From the code in scheduler.rs I understand that this is updating the expired orders. 
If this is correct, I suggest a less broad name for this function. I do not have a good example, but maybe "update_db_periodically"?

gsy-offchain-storage/src/db/trade_service.rs
```diff
@@ -1,54 +1,74 @@
...
-                Ok(document) => {
-                    result.push(document);
-                }
-                _ => {
-                    break;
-                }
-            }
-        }
-        Ok(result)
+        let cursor = self.0.find(doc! {}).await?;
```
hannesdiedrich, line 64: I probably got that wrong from Spyros' earlier PR, but couldn't we pass the `db: &DatabaseWrapper` to the function and then instead of `self.0` use `db`, like in the `init_market_roles` ?

gsy-offchain-storage/src/routes/trades.rs
hannesdiedrich, line 44: @spyrostz Just for my understanding: This is one of the many method that should be called whenever a service is requesting the trades via the EWDS. AFAIS, these are only exposed via HTTP endpoints until now and the next step will be to add EWDS handlers for these, right?

hannesdiedrich (commented):
I have tried to understand the offchain storage part and left quite some questions in order to fully follow. Once These are answered, I will have an easier time to review the rest which is why I pause my review here. Thanks a lot for answering already

gsy-offchain-storage/src/evm_handler.rs
keyanhov: Even with when EWDS_ENABLE_HANDLER is set to true the order are collected from the listener directly from the chain. This to guarantee that order inserted in the offchain storage are indeed the one submitted on the actual onchain contracts

gsy-offchain-storage/src/startup.rs
keyanhov: [here](https://github.com/gridsingularity/gsy-decentralized-exchange/blob/ad7fc4b1ab50e03a1495cd0fadd80d7b420d03b5/gsy-offchain-storage/src/main.rs#L73) at the end of the main.rs

gsy-offchain-storage/src/evm_handler.rs
hannesdiedrich: Thank you, I also realised this in our offline discussion yesterday.

gsy-offchain-storage/src/startup.rs
hannesdiedrich: Ahhh yes, thank you for pointing me to it. IMO a less generic name for this would be beneficial (e.g. `run_http_server`. What do you think?
I can handle this in the refactoring story I am about to grab if you prefer.

gsy-offchain-storage/src/main.rs
hannesdiedrich: In order to be more specific:
In INTELLIGENT all services communicate either via the EWDS or the blockchain. We do not need the http endpoints here, right?

gsy-offchain-storage/src/main.rs
hannesdiedrich: @keyanhov If you agree, I can also do this in the refactoring story.

gsy-offchain-storage/src/routes/trades.rs
hannesdiedrich: maybe this is really also a question for @keyanhov .

gsy-offchain-storage/src/startup.rs
keyanhov: sure, it's fine for me

gsy-offchain-storage/src/main.rs
keyanhov: lgtm

gsy-offchain-storage/src/routes/trades.rs
keyanhov: i'd maintain it for compatibility with the two modes http and ewds. AFAIU, It's still not clear how the execution engine will be integrated in the context of the intelligent project. Please @spyrostz let me know if i miss something wrt to this

keyanhov: i was reading again my reply and perhaps i've not been exhaustive enough in answering to your doubts. [Here](https://github.com/gridsingularity/gsy-decentralized-exchange/blob/feat/GSYE-746/gsy-offchain-storage/src/ewds_handler.rs#L358) you can find the implementation for the EWDS handler for the trades query, which has already been implemented. Even if it's not clear if we'll need to modify/update this once we'll have clear integration plan for INTELLIGENT

gsy-contracts/contracts/GsyVault.sol
```diff
@@ -0,0 +1,176 @@
...
+        bytes16 indexed actorId,
+        address indexed wallet,
+        uint256 amount
+    );
+    event TransferredBySettlement(
+        bytes16 indexed fromActorId,
+        bytes16 indexed toActorId,
+        uint256 amount
+    );
+    event ProxyUpdated(
```
spyrostz, line 43: Nothing to do here, just a question though: the proxy functionality was added here only to be proactive and because it does not hurt, or is there a usecase that you envision ?

gsy-contracts/contracts/OrderRegistry.sol
```diff
@@ -0,0 +1,148 @@
...
+        uint64 energy;
+        uint64 energyRate;
+        bool isBid;
+    }
+
+    // Intelligent Order UUID (bytes16) => Status
+    mapping(bytes16 => OrderStatus) public orderStatus;
+    mapping(bytes16 => OrderParams) private orders;
+
+    event OrderPlaced(
```
spyrostz, line 40: This is missing the energy source preference and energy type (which are useful for user preferences). IMO these parameters should be added to the contract too. If you agree to add them, then let me know to add a separate ticket for this

gsy-contracts/contracts/TradeSettlement.sol
```diff
@@ -0,0 +1,195 @@
...
+    bytes32 public constant EXECUTION_ENGINE_ROLE =
+        keccak256("EXECUTION_ENGINE_ROLE");
+
+    OrderRegistry public registry;
+    GsyVault public vault;
+
+    event TradeSettled(
+        bytes16 indexed tradeId,
+        bytes16 indexed bidId,
+        bytes16 indexed askId,
```
spyrostz, line 23: Better to remain consistent and name it as "offerId"

gsy-contracts/contracts/TradeSettlement.sol
```diff
@@ -0,0 +1,195 @@
...
+        if (
+            trade.selectedEnergy > trade.bid.energy ||
+            trade.selectedEnergy > trade.ask.energy
+        ) {
+            revert EnergyMismatch();
+        }
+
+        uint256 totalCost = trade.selectedEnergy * trade.clearingPrice;
+
+        vault.transferBySettlement(
```
spyrostz, line 159: IMO this should be omitted. Billing is not managed internally, but by the billing and payment service which is completely separate from the DEX.

gsy-contracts/contracts/TradeSettlement.sol
```diff
@@ -0,0 +1,195 @@
...
+            trade.bid.createdBy,
+            trade.ask.createdBy,
+            totalCost
+        );
+
+        registry.updateStatus(trade.bid.orderId, OrderRegistry.OrderStatus.Executed);
+        registry.updateStatus(trade.ask.orderId, OrderRegistry.OrderStatus.Executed);
+
+        emit TradeSettled(
```
spyrostz, line 168: IMO this event should include all necessary information for the creation of an offchain Trade object, even if it bloats the transaction storage. As far as I see, buyer, seller, time_slot should also be added. More importantly residual_bid_id, residual_offer_id is very important to be added since they cannot be deduced by the offchain storage.

gsy-contracts/scripts/deploy.ts
```diff
@@ -0,0 +1,149 @@
...
+  );
+  await orderRegistry.waitForDeployment();
+  const orderRegistryAddress = await orderRegistry.getAddress();
+
+  const tradeSettlementFactory =
+    await ethers.getContractFactory("TradeSettlement");
+  const tradeSettlement = await tradeSettlementFactory.deploy(
+    orderRegistryAddress,
+    gsyVaultAddress,
+  );
```
spyrostz, line 87: Minor, but since the code is duplicated it makes sense to refactor to a single function that is called for each contract, and returns the deploy() return value and the contract address as a tuple.

gsy-contracts/test/TradeSettlement.ts
```diff
@@ -0,0 +1,256 @@
+import { loadFixture } from "@nomicfoundation/hardhat-toolbox/network-helpers";
```
spyrostz, line 1: General comment for the tests and related to one of my former comments, I a test for residual trades is missing, that would more or less simulate a partially filled order creation. Could be done as an e2e test, but IMO it would be nice to have an isolated test here for this (common) usecase

gsy-ethers-listener/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        let mut stream_order_placed = order_placed_filter.subscribe().await?;
+        let mut stream_order_cancelled = order_cancelled_filter.subscribe().await?;
+        let mut stream_trade_settled = trade_settled_filter.subscribe().await?;
+        let mut stream_market_status = market_status_filter.subscribe().await?;
+
+        info!("GSy Ethers Listener started. Waiting for events...");
+
+        loop {
+            tokio::select! {
```
spyrostz, line 73: I know that the docker container is supposed to restart the service if the network connection / websocket fails, however I think it is worth it to add a reconnection process here, so that the service can withstand minor network interruptions.

gsy-ethers-listener/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+                    match log {
+                        Ok(event) => {
+                            info!("Detected OrderPlaced: {:?}", hex::encode(event.order_id));
+                            if let Err(e) = self.handler.handle_order_placed(event).await {
+                                error!("Error handling OrderPlaced: {:?}", e);
+                            }
+                        },
+                        Err(e) => error!("Error in OrderPlaced stream: {:?}", e),
+                    }
+                }
```
spyrostz, line 84: This is also repetitive, however it will be hard to refactor and maintain the logging as is, therefore feel free to leave it as is .

gsy-ethers-listener/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        let provider = Provider::<Ws>::connect(&self.config.node_url).await?;
+        let client = Arc::new(provider);
+
+        let order_registry = GsyContracts::new(self.config.order_registry_address, client.clone());
+        let trade_settlement =
+            GsyContracts::new(self.config.trade_settlement_address, client.clone());
+        let market_controller =
+            GsyContracts::new(self.config.market_controller_address, client.clone());
+
+        let order_placed_filter = order_registry.event::<OrderPlacedFilter>();
```
spyrostz, line 60: Just for my sake, can you please add a comment that these types (OrderPlacedFilter, OrderCancelledFilter ... ) are automatically deduced by the contract ABI ? Because I am sure that I will forget. Thanks in advance!

gsy-execution-engine/src/connectors/evm_connector.rs
```diff
@@ -0,0 +1,176 @@
...
+use anyhow::{anyhow, Result};
+use ethers::prelude::*;
+use ethers::utils::keccak256;
+use gsy_offchain_primitives::utils::parse_or_hash_bytes16;
+use std::{str::FromStr, sync::Arc};
+use tracing::{info, warn};
+
+abigen!(
+    TradeSettlementContract,
+    r#"[
```
spyrostz: I think I would prefer the contract definition to be isolated in its own module, and to be imported here.

gsy-contracts/contracts/GsyVault.sol
keyanhov: This is something we also had on the substrate version where we enabled actor wallet delegation (some delegate can insert order on behave of another user we delegate his order submission). The registrar maps the Actor UUID to an authorized wallet, and setProxy lets that actor authorize a service/operator/delegate wallet without changing the Actor UUID mapping. This can be useful if an aggregator/service needs to submit or operate on behalf of an actor. If we want to keep this version strictly limited to direct actor wallets, we can remove it, but I kept it to be consistent with the design we implemented in the substrate version

gsy-contracts/contracts/OrderRegistry.sol
keyanhov: I agree these fields are useful domain information. I would keep this as a separate ticket though, because adding them changes OrderParams, the OrderPlaced ABI/event, and all event consumers. In the current path the contract only validates settlement-critical order data, while energy type/source preferences remain matching-engine layer inputs. If we want them enforceable or indexable on-chain, we should add them in a dedicated ABI-change ticket

gsy-contracts/contracts/TradeSettlement.sol
keyanhov: Agreed, i'm going to update this in next commit to address this and other comments

gsy-contracts/contracts/TradeSettlement.sol
keyanhov: I agree and i'm going to update this in the next commit. Given this change, should we also remove the collateral/deposit/withdrawal responsibility from GsyVault? At that point the contract is no longer really a vault it is mostly an authorization registry. If that matches the intended design, I’d suggest renaming it in a follow-up to something like ActorRegistry and removing the balance/deposit/withdraw path as well

gsy-contracts/contracts/TradeSettlement.sol
keyanhov: Agreed. I'm going to update this in the next commit

gsy-contracts/scripts/deploy.ts
keyanhov: Agreed. I'll add a deployContract(...) helper that returns the deployed contract and address tuple

gsy-contracts/test/TradeSettlement.ts
keyanhov: 👍 I'll add an isolated partial-fill test hat emits and checks the residual offer ID in the TradeSettled event

gsy-ethers-listener/src/lib.rs
keyanhov: 👍 i'm thinking to add a reconnection loop in the next commit

gsy-contracts/contracts/GsyVault.sol
spyrostz: Agreed and let's keep it this way. Thanks for the clarification!

gsy-contracts/contracts/TradeSettlement.sol
spyrostz: Indeed this should be changed and I agree that the collateral / deposit / withdrawal functionality is better to be removed, to not increase the blockchain size. I will create a separate ticket for this to be done in isolation, thanks!

gsy-contracts/contracts/OrderRegistry.sol
spyrostz: Agreed and created a ticket for this https://gridsingularity.atlassian.net/browse/DD-399

gsy-contracts/contracts/TradeSettlement.sol
spyrostz: Should be done in this ticket https://gridsingularity.atlassian.net/browse/DD-399

gsy-contracts/contracts/TradeSettlement.sol
spyrostz: Created a ticket for this https://gridsingularity.atlassian.net/browse/DD-400

gsy-contracts/contracts/TradeSettlement.sol
spyrostz: Should be done in this ticket https://gridsingularity.atlassian.net/browse/DD-399

gsy-market-orchestrator/src/orchestrator.rs
```diff
@@ -1,90 +1,280 @@
...
+    delivery_time: u64,
+    open_offset_mins: i64,
+    close_offset_mins: i64,
+) -> bool {
+    let open_time = (delivery_time as i64 + open_offset_mins * 60) as u64;
+    let close_time = (delivery_time as i64 + close_offset_mins * 60) as u64;
+    now >= open_time && now < close_time
+}
+
+#[cfg(test)]
```
hannesdiedrich, line 128: @keyanhov IMO we should homogenize the place where we put the tests.
For my personal taste, these belong into the tests folder. 
Would you agree or do you have a strong opinion against it?

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -0,0 +1,675 @@
...
+    start_time: Option<u64>,
+    #[serde(alias = "endTime")]
+    #[serde(default)]
+    end_time: Option<u64>,
+    #[serde(alias = "areaUuid")]
+    #[serde(default)]
+    area_uuid: Option<String>,
+}
+
+pub async fn start_ewds_request_handler(db: DatabaseWrapper, config: EwdsHandlerConfig) {
```
hannesdiedrich, line 191: I think the connection to the EWDS is not tested, or can you please point me to it?
I would expect a test that sends a request to the EWDS and this handler should react.
Is it a problem to mock the EWDS connection?

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -0,0 +1,675 @@
...
+}
+
+async fn handle_request(
+    db: &DatabaseWrapper,
+    client: &Client,
+    config: &EwdsHandlerConfig,
+    envelope: EwdsRequestEnvelope,
+) -> Result<()> {
+    match envelope.operation.as_str() {
+        "orders.query" => {
```
hannesdiedrich, line 326: NIT and nothing to do now, but how about using enums for `orders.query`, `trades.query` etc.?

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -0,0 +1,675 @@
...
+struct EwdsRequestEnvelope {
+    #[serde(alias = "request_id")]
+    request_id: String,
+    operation: String,
+    payload: Value,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct EwdsOrderDto {
```
hannesdiedrich, line 107: These seem to be duplicated (I found them also in the matching engine).
Should we move them to the primitives and re-use in multiple places?
I could add this to my refactoring story.

gsy-market-orchestrator/src/orchestrator.rs
keyanhov: I don’t have a strong opinion against moving tests, but in this specific case the tests exercise private orchestration helpers and pure functions. Keeping them inline avoids exposing internals only for tests. I’d keep these as module-level unit tests (also a rust lang convention), and use tests/ folder for public integration behavior

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: You’re right, we currently do not have an automated mocked-CGW test for this handler. The flow was validated with the e2e tests. We may need a small refactor to expose a one-shot process_batch/handler path for tests instead of testing the infinite loop directly. Wdyt?

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: Yes we can use enum here. I’d treat this as a small cleanup/refactor  (i.e. EwdsOperation::{OrdersQuery, TradesQuery, MeasurementsQuery}) with serde renames for orders.query, trades.query, etc.. I'd handle this in the context of a separate specific ticket if you are okay with this. @spyrostz tagging also you for visibility and evaluating the priorities

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: Yes, you're correct. I’d include this in another ticket to avoid expanding the current PR further, unless we decide this duplication is worth fixing immediately

gsy-market-orchestrator/src/orchestrator.rs
hannesdiedrich: > Keeping them inline avoids exposing internals only for tests. 
I just learned that when creating tests in my own branch. So I agree on leaving them here for now. Thanks

gsy-offchain-storage/src/ewds_handler.rs
hannesdiedrich: yes, with the help of AI I locally created a test that tests the private `handle_request`, which was the reason for my comment here. Lets mark this conversation as done and let me add the test in my branch.

gsy-offchain-storage/src/ewds_handler.rs
hannesdiedrich: Yes, definitely nothing for now. Thanks for dealing with this later.

gsy-offchain-storage/src/ewds_handler.rs
hannesdiedrich: IMO this is not important now, I just wanted to rise a flag here. Let's change this in a future ticket (I might also add this to my branch because I am adding a match case here.

keyanhov: works good for me. I'd propose to track these changes and the changes required [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/46#discussion_r3426780216) in the same ticket

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -37,16 +52,322 @@ pub async fn fetch_trades_and_measurements_for_timeslot(
...
+        "measurementsQueryResponse",
+    )
+    .await?;
 
     Ok((trades, measurements))
-}
\ No newline at end of file
+}
+
+async fn query_via_ewds<T: DeserializeOwned>(
```
hannesdiedrich, line 199: This method seems to be quite universal and could be re-used in other services. Should we put this in the primitives?

keyanhov: definitely, i'd track/tackle this in the above mentioned ticket where we can describe and track the primitives and offchain-primitives structs and fn

gsy-ethers-listener/src/lib.rs
hannesdiedrich: I like to ask here, why are the events called *Filter here but not in the contracts?
I see `emit MarketStatusUpdated(..` in the contract, but here it is `MarketStatusUpdatedFilter`

gsy-offchain-storage/src/ewds_handler.rs
spyrostz: Thanks and ticket added https://gridsingularity.atlassian.net/browse/DD-401

gsy-offchain-storage/src/ewds_handler.rs
spyrostz: https://gridsingularity.atlassian.net/browse/DD-401

gsy-execution-engine/src/connectors/offchain_storage.rs
spyrostz: I would suggest to first move these in the offchain primitives in the context of this ticket https://gridsingularity.atlassian.net/browse/DD-401. I think that it is preferable to merge the 2 crates afterwards https://gridsingularity.atlassian.net/browse/DD-402.

gsy-market-orchestrator/src/orchestrator.rs
```diff
@@ -1,90 +1,280 @@
...
-	buffer.extend_from_slice(&delivery_timestamp.to_be_bytes());
-	H256(blake2b(32, &[], &buffer).as_bytes().try_into().expect("hash is 32 bytes"))
+async fn orchestrate_markets_at<C>(config: &Config, client: &C, now: u64) -> anyhow::Result<()>
+where
+    C: MarketChainClient + ?Sized,
+{
+    let look_ahead_horizon = now + (config.look_ahead_hours * 3600);
+
+    let mut current_delivery_secs =
+        (now / GLOBAL_CONSTANTS.time_slot_sec) * GLOBAL_CONSTANTS.time_slot_sec;
```
hannesdiedrich, line 63: I think there is a bug here and I am not sure what you were trying to calculate. 
I will probably re-write this part, but can you please explain to me why you are not using `now` directly?

hannesdiedrich: Did you mean to calculate the start of the market slot?
```rust
let mut current_delivery_secs = now - (now % GLOBAL_CONSTANTS.time_slot_sec);
```
Also @spyrostz: how about calling `orchestrate_markets_at` only every 15 minutes? Or is this too risky?
I am asking, because we can reduce the load on the blockchain and DB if we do not call it every tick as it is now.

spyrostz: Read this quickly and will check it out in more detail later, but this is clamping the current time to the last time slot. I think that these are all integer values, and integer division removes the floating point part. Same for the multiplication. This achieves to calculate the most recent time_slot (essentially the start of the current delivery slot). Take all these with a grain of salt though, I will read it later in more detail to see where this fits in the entire PR.

hannesdiedrich: Ah, i was fearing that this is something implicit to the types. Makes sense now.

gsy-offchain-storage/src/db/market_service.rs
```diff
@@ -4,51 +4,54 @@ use futures::StreamExt;
...
-        Ok(result)
-    }
-
-    #[tracing::instrument(name = "Fetching markets for a community", skip(self))]
-    pub async fn get_community_market(
+    #[tracing::instrument(name = "Fetching markets", skip(self))]
+    pub async fn filter(
         &self,
-        community_id: String,
+        market_id: Option<String>,
```
hannesdiedrich, line 43: IMO the `market_id` should not be part of the input parameters here. As this value is (must be) unique, there is no point in filtering for it. A getter function should deal with it.
Why do I mention it here? I would like to add another filter parameter here: `market_type` in order to filter the markets for specific market types when requesting from the orchestrator side. 
Do you agree?

offchain-primitives/src/db_api_schema/market.rs
```diff
@@ -18,3 +37,55 @@ pub struct MarketSchema {
...
+
+#[derive(Serialize, Deserialize, Debug, Encode, Decode, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct IntelligentMarketSchema {
+    pub market_id: String,
+    pub community_id: String,
+    pub opening_time: String,
+    pub closing_time: String,
+    pub delivery_start_time: String,
+    pub delivery_end_time: String,
```
hannesdiedrich, line 66: shouldn't we save all these times as u32 just as in MarketTopologySchema in order to not having the problem of converting between string and unix time in the process?

offchain-primitives/src/db_api_schema/market.rs
```diff
@@ -18,3 +37,55 @@ pub struct MarketSchema {
...
+#[derive(Serialize, Deserialize, Debug, Encode, Decode, Clone, PartialEq, Eq)]
+pub enum MatchingAlgorithm {
+    PayAsBid,
+    PayAsClear,
+    AMM,
+}
+
+#[derive(Serialize, Deserialize, Debug, Encode, Decode, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct IntelligentMarketSchema {
```
hannesdiedrich, line 60: Why did you add a new one? I think the MarketSchema is already made for Intelligent (only needed to be adapted)

offchain-primitives/src/db_api_schema/market.rs
```diff
@@ -18,3 +37,55 @@ pub struct MarketSchema {
     pub delivery_end_time: String,
     pub market_type: MarketType,
 }
+
+#[derive(Serialize, Deserialize, Debug, Encode, Decode, Clone, PartialEq, Eq)]
+pub enum IntelligentMarketType {
+    #[serde(rename = "local_spot")]
```
hannesdiedrich, line 43: I understand that you can create convert from enum to string like:
```rust
let mt: IntelligentMarketType = serde_plain::from("flex")?;
let s = serde_plain::to_string(&mt)?;  
```
Right?
Shouldn't we also add this functionally to the MatchingAlgorithm?

gsy-matching-engine/src/connectors/redis_connector.rs
```diff
@@ -25,10 +25,12 @@ pub fn value_to_f32(value: &Value) -> f32 {
 pub fn value_to_datetime(value: &Value) -> Option<NaiveDateTime> {
     // Helper function to convert the serde Value to NaiveDateTime
     match value.as_str() {
-        Some(..) => match NaiveDateTime::parse_from_str(value.as_str().unwrap(), "%Y-%m-%dT%H:%M:%S") {
-            Ok(datetime) => Some(datetime),
-            Err(_e) => None,
-        },
+        Some(..) => {
```
spyrostz, line 20: I think it is time to fix these formatting errors in this branch too. Nothing to do here, opened a separate ticket to deal with this issue after this PR is merged https://gridsingularity.atlassian.net/browse/DD-403

gsy-matching-engine/src/primitives/web2.rs
```diff
@@ -1,26 +1,47 @@
...
             serializer.serialize_str(&s)
-        },
-        None => serializer.serialize_none()
+        }
+        None => serializer.serialize_none(),
     }
 }
 
+#[derive(Clone, Debug, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub enum EnergyType {
```
spyrostz: This and the requirements/attributes structs should be part of the orders.rs module in offchain-primitives even now IMO

gsy-ethers-listener/src/lib.rs
keyanhov: addressed [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/53)

gsy-ethers-listener/src/lib.rs
keyanhov: That `*Filter` suffix is not a contract naming choice. The Solidity event is still named `MarketStatusUpdated`. In Rust, the type `MarketStatusUpdatedFilter` is generated by `ethers::abigen!` from the event ABI. `ethers-rs` names generated event structs with the `Filter` suffix because they are used both to build event filters/subscriptions and to decode matching log entries.

So:

Solidity:
`event MarketStatusUpdated(...)`

Rust generated by `abigen!`:
`MarketStatusUpdatedFilter`

They refer to the same on-chain event. The suffix is just an ethers-rs codegen convention.

docs/setup/docker.md
hannesdiedrich: IMO still valid, but can be done later.

gsy-offchain-storage/src/startup.rs
hannesdiedrich: Done in https://github.com/gridsingularity/gsy-decentralized-exchange/pull/50

gsy-offchain-storage/src/main.rs
hannesdiedrich: Done in here https://github.com/gridsingularity/gsy-decentralized-exchange/pull/50

gsy-ethers-listener/src/lib.rs
hannesdiedrich: Ah, thank you for explaining again, I should have resolved this comment already. 
However, I agree with Spyros to leave an inline comment in order to not ask again in the future.

gsy-offchain-storage/src/db/market_service.rs
spyrostz: You are thinking this from the orchestrator side only though. `market_id` for example is useful for the EMS to fetch information for a specific market, for example to check if the market is open or closed. I would recommend to maintain the optional market_id parameter and I agree to add an additional `market_type` parameter as you said to accomodate the orchestrator with the same method. This is better to be dealt with in a separate ticket or in the orchestrator ticket IMO. Feel free to open a separate ticket if you think it is better to be dealt with separately.

gsy-offchain-storage/src/db/market_service.rs
```diff
@@ -74,17 +77,22 @@ impl MarketService {
         Ok(result)
     }
 
-    #[tracing::instrument(
-        name = "Saving market to database",
-        skip(self, market),
-        fields(market = ?market)
-    )]
+    #[tracing::instrument(name = "Saving market", skip(self, market), fields(market = ?market))]
     pub async fn insert(&self, market: MarketSchema) -> Result<MarketSchema> {
```
spyrostz: Very minor, but maybe it makes sense to rename the method to `upsert` now

gsy-offchain-storage/src/db/order_service.rs
```diff
@@ -80,22 +99,25 @@ impl OrderService {
     pub async fn filter_orders(
         &self,
         market_id: Option<String>,
-        start_time: Option<String>,
-        end_time: Option<String>,
+        start_time: Option<u64>,
```
spyrostz, line 102: Out of curiosity, is the conversion to unsigned preferrable for indexing ? Because IIRC the type that we are supposed to receive from EWDS is an ISO datetime string. At any case I will open a followup ticket that will check the orders / trade endpoints with the ontology (these were defined by us, TUM and UoC and not the pilot sites, therefore we have to adhere to them already)

spyrostz (commented):
Added some comments, only one to act on

gsy-offchain-storage/src/db/order_service.rs
spyrostz: Added ticket so resolving the comment https://gridsingularity.atlassian.net/browse/DD-406

gsy-offchain-storage/src/db/market_service.rs
hannesdiedrich: I only learn now, that there seems to be no real `get()` method in the mongo Interface and that you always have to call `filter()` even if you know a unique property. 
So, indeed, we can keep this, even if we already have `get_market()` that only applies the filter for the market_id

gsy-matching-engine/src/primitives/web2.rs
```diff
@@ -22,34 +25,15 @@ where
...
 
 #[derive(Clone, Debug, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Bid {
     pub r#type: String,
     pub id: String,
     pub market_id: String,
-    pub energy: f32,
-    pub energy_rate: f32,
-    pub original_price: f32,
+    pub energy: f64,
```
hannesdiedrich, line 34: I can not find the original comment that asks for this type change. Can you please shortly explain why you went for the "larger" type?

hannesdiedrich (approved):
LGTM

gsy-matching-engine/src/primitives/web2.rs
spyrostz: Because there was a type mismatch between these types. The Order type from the primitives used f64 while the type in the matching engine uses f32. There is literally no need for a different type since memory is enough for all processes, and floating point types are not used on-chain.

PR #47, opened 2026-06-04
Ontology update

spyrostz:


gsy-offchain-storage/openapi.yaml
```diff
@@ -0,0 +1,999 @@
+openapi: 3.1.0
```
hannesdiedrich, line 1: Did you create this file manually or with a tool?

gsy-offchain-storage/src/event_listener.rs
```diff
@@ -1,80 +1,7 @@
...
-                    Err(e) => {
-                        tracing::error!("Failed to execute update: {:?}", e);
-                    }
-                }
-            }
-        }
-    }
 
+pub async fn init_event_listener(_db: DbRef, _node_url: String) -> Result<(), Error> {
+    // TODO: Integrate with Ethereum Event Listener
```
hannesdiedrich, line 5: Sorry to ask again, but why to we need a connection to the DB here for this project?

gsy-offchain-storage/src/routes/trades.rs
```diff
@@ -45,16 +14,19 @@ pub async fn post_normalized_trades(trades: Json<Vec<TradeSchema>>, db: DbRef) -
...
 pub struct GetTradesParams {
-    start_time: Option<u32>,
-    end_time: Option<u32>,
+    start_time: Option<String>,
+    end_time: Option<String>,
 }
 
 #[tracing::instrument(name = "Retrieve trades", skip(db))]
 pub async fn get_trades(db: DbRef, query_params: Query<GetTradesParams>) -> impl Responder {
     match db
```
hannesdiedrich, line 27: For trying around I added the following code block to this method:
```rust
    let fmt = "%Y-%m-%dT%H:%M:%S";
    let start = NaiveDateTime::parse_from_str(&query_params.start_time.clone().unwrap(), fmt);
    let end   = NaiveDateTime::parse_from_str(&query_params.end_time.clone().unwrap(), fmt);

    match (start, end) {
        (Ok(start), Ok(end)) => {
            if end < start {                      // now both are NaiveDateTime
                return HttpResponse::BadRequest()
                    .body("end_time must be after start_time");
            }
            // proceed — start and end are valid NaiveDateTime here
        }
        _ => {
            return HttpResponse::BadRequest()
                .body("start_time and end_time must be valid datetimes");
        }
    }
```
Is this validation done somewhere or should we leave it to the client to provide the correct datetimes?
I understand that this is probably an overkill, right?

gsy-offchain-storage/src/routes/trades.rs
```diff
@@ -2,39 +2,8 @@ use crate::db::DbRef;
...
 use serde::Deserialize;
 
-#[tracing::instrument(
-    name = "Adding new trades",
-    skip(trades, db),
-    fields(
-    trades = ?trades
-    )
-)]
-pub async fn post_trades(trades: Json<Vec<u8>>, db: DbRef) -> impl Responder {
```
hannesdiedrich, line 15: This is one task for the to be implemented service that listens to the client gateway, correct?

gsy-offchain-storage/src/db/grid_topology_service.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+    Ok(result)
+}
+
+#[repr(transparent)]
+pub struct AssetService(pub Collection<AssetSchema>);
+
+impl AssetService {
+    pub async fn insert_assets(&self, assets: Vec<AssetSchema>) -> Result<HashMap<usize, Bson>> {
+        Ok(self.0.insert_many(assets).await?.inserted_ids)
```
hannesdiedrich, line 102: this self.0 is making my head hurt. Also LLM did not help much.
I understand that `AssetService` is a implementation of `AssetSchema`, correct?
Self.0 returns the first element of which tuple?
I understand that `.insert_many` is a db operator but i fail to connect the wires here.
Can you please help me?

hannesdiedrich (commented):
I left some questions. Will LGTM ones answered. Thanks!

gsy-offchain-storage/openapi.yaml
spyrostz: Created this automatically with Claude

gsy-offchain-storage/src/event_listener.rs
spyrostz: This is not a connection to the DB. This function reads events from the blockchain, for example an order creation or trade creation event, and record it off-chain.

gsy-offchain-storage/src/routes/trades.rs
spyrostz: No, this is actually an old endpoint from Fedecom, which received the trades directly from the off-chain worker, as a series of bytes (Vec<u8>). This was then decoded using SCALE codec to the Trade struct. All these are not needed anymore though, because there is no offchain worker and these objects are read via blockchain events (as per the comment above)

gsy-offchain-storage/src/db/grid_topology_service.rs
spyrostz: I think that we discussed about this offline, this is a Collection type, similar to Django's ModelManager class.

gsy-offchain-storage/src/routes/trades.rs
spyrostz: Refactored and modified a bit the validation to not use the custom format. Thanks for the suggestion!

hannesdiedrich (approved):
LGTM

PR #49, opened 2026-06-12
Fedecom deployment

spyrostz:


e2e-tests/src/steps/cross_market_steps.rs
```diff
@@ -0,0 +1,357 @@
...
+				same_market_trades += 1;
+				info!(
+					"Same-market trade settled in {:?} ({} energy units)",
+					trade.market_id, trade.parameters.selected_energy
+				);
+			}
+		}
+	}
+
+	assert!(
```
keyanhov: The test prepares expected_min = world.cross_communities.len() at line 247 and only returns early once same_market_trades >= expected_min at line 263, but after the loop it accepts any positive number of trades. That means a scenario with two community markets can pass after observing only one same-market trade

keyanhov (commented):
Market isolation is enforced by the matching engine, but not by the settlement pallet. `validate` checks energy, rate, and time slot, but it never verifies that `bid_offer_match.market_id`, `bid.bid_component.market_id`, and `offer.offer_component.market_id` are equal. A registered operator can call settle_trades directly with bid and offer from different markets as long as the time slots match

e2e-tests/src/steps/cross_market_steps.rs
spyrostz: Corrected the test to catch this cornercase.

PR #50, opened 2026-06-16
Feat/gsye 397

hannesdiedrich:


gsy-offchain-storage/src/configuration.rs
```diff
@@ -41,8 +41,8 @@ impl Settings {
             self.database_host
         )
     }
-    pub fn get_scheduler_interval(&self) -> u32 {
-        self.scheduler_interval
+    pub fn get_update_db_interval(&self) -> u32 {
```
spyrostz: I think it is better to leave as is, since the update_db term is misleading. Update DB is not only done by the DB scheduler task (which checks for expired orders) but also by normal HTTP requests and other jobs such as EWDS handler or the blockchain event listener. Maybe can be renamed to `get_db_scheduler_interval` or similar, but I consider it minor.

gsy-offchain-storage/src/update_db.rs
```diff
@@ -3,8 +3,9 @@ use chrono::Utc;
 use gsy_offchain_primitives::db_api_schema::orders::OrderStatus;
 use tokio_schedule::{every, Job};
 
-pub async fn start_scheduler(db: DbRef, scheduler_interval: u32) {
-    let every_interval = every(scheduler_interval)
+/// Periodically mark stale open orders as `Expired` using `time_slot` and current time
+pub async fn update_db_periodically(db: DbRef, update_interval: u32) {
```
spyrostz: Same as above, maybe can be made more explicit by naming it as `expire_orders_scheduler`

spyrostz (approved):
2 naming comments, LGTM already

gsy-offchain-storage/src/configuration.rs
hannesdiedrich: I see that this method is only used  once in the offchain storage main.rs.
```
let scheduler_interval = configuration.get_scheduler_interval();
```
Why is this method needed, can't we access the Settings directly with:
```rurst
let scheduler_interval = configuration.update_interval.clone();
```
?

gsy-offchain-storage/src/update_db.rs
hannesdiedrich: Yes, this is a better name. Thanks!

gsy-offchain-storage/src/configuration.rs
spyrostz: We can and I do not disagree since scheduler_interval has public access, however this does not change the main issue here. `update_interval` is a very generic name for this, and I would recommend to maintain the `scheduler_interval` name, or edit it to reflect that this job scheduler only updates expired orders.

PR #53, opened 2026-06-23
Feat/dd 396

keyanhov:


gsy-contracts/scripts/gas-report.ts
```diff
@@ -0,0 +1,682 @@
...
+    },
+    residualBidId: ethers.ZeroHash.slice(0, 34),
+    residualOfferId,
+    selectedEnergy: 100_000,
+    clearingPrice: 12_000,
+  };
+
+  await recordTx(
+    "Mutating calls",
+    "settleBatch(Match[1])",
```
spyrostz: Batch size is too small. I think it makes sense to add more dummy bids and offers, and call the settleBatch function with more inputs to estimate the effect of the batch size on the gas cost.

spyrostz (commented):
Some comments mostly related to the batch size of the contract calls. Will LGTM once fixed

gsy-contracts/scripts/gas-report.ts
keyanhov: 💯completely agree. I'll change the script accordingly

gsy-contracts/scripts/gas-report.ts
```diff
@@ -0,0 +1,758 @@
...
+    "Mutating calls",
+    "updateStatus(bytes16,OrderStatus)",
+    "OrderRegistry",
+    orderRegistryContract.updateStatus(statusOrderId, 2),
+    "Measured with benchmark-only settlement role granted to deployer.",
+  );
+
+  const settleBatchMatches = new Map<number, any[]>();
+  settleBatchMatches.set(1, [
+    buildMatch(tradeId, bidOrder, offerOrder, residualOfferId),
```
spyrostz, line 612: I would probably like to have multiple bids and offers. At the moment still the batch size is 1, and the settleBatch function would only be called with a single match. Could there not be multiple bids and offers generated in a loop, that lead to multiple matches ?

spyrostz (commented):
One comment with regards to batch size. Will LGTM once resolved.

gsy-contracts/scripts/gas-report.ts
keyanhov: The line you pointed at is only the baseline Match[1] case. Right below it, the script iterates over the configured batch sizes except 1, creates N unique bid/offer pairs, builds N matches, and then records settleBatch(Match[N]). The default is GAS_REPORT_SETTLE_BATCH_SIZES=1,2,5, so the report includes single-match and multi-match settlement rows. The batch sizes can also be overridden via env var, e.g. GAS_REPORT_SETTLE_BATCH_SIZES=1,5,10,25, if we want to benchmark larger batches.

PR #54, opened 2026-06-25
Homogenize schemas

hannesdiedrich:


gsy-market-orchestrator/src/config.rs
```diff
@@ -49,7 +51,7 @@ pub struct MarketRule {
 pub static MARKET_RULES: Lazy<Vec<MarketRule>> = Lazy::new(|| {
     vec![
         MarketRule {
-            market_type: MarketType::Spot,
+            market_type: MarketType::LocalSpot,
```
spyrostz: This is my bad because I did not see the enum values when reviewing the ontology. This is debatable to be frank, because in future projects we associate the naming with the one in the INTELLIGENT ontology. IMO it is preferrable to keep the original Spot Flexibility and Settlement names and change the ontology instead. I doubt that any partner will have issues with this change. I will do the ontology coordination, but if you revert this change it would be great .

gsy-offchain-storage/tests/api/market.rs
```diff
@@ -1,264 +0,0 @@
-use crate::helpers::{init_app, stop_app};
```
spyrostz, line 1: These unit tests should be included and fixed IMO. For example, I see that the ClearingResult is also tested here, and the endpoints for it are not removed if I saw correctly. Also some of the market tests could stay, e.g. get_market_succeeds and get_market_from_community_succeeds

offchain-primitives/src/utils/mod.rs
```diff
@@ -99,3 +100,12 @@ pub fn read_env_or<T: FromStr>(variable_name: &str, default_value: T) -> T {
         .and_then(|s| s.parse().ok())
         .unwrap_or(default_value)
 }
+
+pub fn format_timeseries_timestamp(timestamp: u64) -> String {
```
spyrostz: I think that this naming is rather misleading, because timestamp formatting implies that a datetime string is the output of it. I would prefer `timestamp_to_string_with_padding`

hannesdiedrich: I totally agree, just kept it from GSYE-746

gsy-offchain-storage/tests/api/market.rs
hannesdiedrich: ups, let me check and re-add again

gsy-market-orchestrator/src/config.rs
hannesdiedrich: I totally agree and am happy that you suggest that.

gsy-offchain-storage/tests/api/market.rs
hannesdiedrich: I re-added them again, but I removed an endpoint that is not needed any more IMO in this commit:
[9a3b1c7](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/54/commits/9a3b1c7ad38e16c26b8426ab26a329e86f9d7b04)

spyrostz (approved):
LGTM

PR #58, opened 2026-07-02
Refactor/dd 401

keyanhov:


hannesdiedrich (approved):
Thank you!
LGTM

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -169,201 +126,31 @@ async fn fetch_trades_and_measurements_via_ewds(
...
-    )
+    let measurements: Vec<MeasurementSchema> = query_via_ewds(EwdsQueryRequest {
+        operation: EwdsOperation::MeasurementsQuery,
+        query_payload: query,
+        request_topic_env: "EWDS_MEASUREMENTS_REQUEST_TOPIC",
+        request_topic_default: "measurementsQuery",
+        response_topic_env: "EWDS_MEASUREMENTS_RESPONSE_TOPIC",
+        response_topic_default: "measurementsQueryResponse",
+        response_client_id_env: "EWDS_EXECUTION_ENGINE_CLIENT_ID",
+        response_client_id_default: "gsyexecutionengine",
```
hannesdiedrich: I have a genera question here. Why do we care so much about the default values here? Or rather: why do we want to be able to control the channel names via environmental variables? I guess that these names are changed while we run the dex, right? But is this really a possible scenario? 
IMO: YAGNI and the code would look much simpler if we put the channel names into enums and re-use here when constructing EwdsQueryRequest. If a channel name changes, we change the enum and re-deploy. What do you all think?

keyanhov: Using environment variables keeps the topic names deployment-configurable without requiring a code change and rebuild. This is useful when different environments, EWDS instances, or migration periods use different naming conventions. The defaults preserve the standard setup, while overrides keep the connector reusable and avoid coupling deployment configuration to the binary

hannesdiedrich: I understand that, but IMO rebuilding has the same complexity as restarting the docker container with the different environmental variables (if you ignore the waiting time for the build). Even though I do not thing that these channel names will ever change in the frame of the project I am fine to read the environmental variables for this. It is just that the interface of `query_via_ewds` is bloated unnecessarily (IMO). Why don't we ignore the default values and assume that all of these env variables are set? If not, the method should write an error log. This will give the maintainer the ability to recognise and to set the variable. 
What do you think?

keyanhov: IMO, rebuilding is noticeably more expensive than changing an environment variable and restarting the container. For that reason, I think keeping the topic names configurable at deployment time still has value. That said, I agree the current `query_via_ewds` interface is more verbose than necessary we could move the topic resolution (including defaults) into the EWDS layer and keep the caller much simpler

keyanhov: Also, we already have a configuration layer in gsy-offchain-storage that resolves defaults and environment overrides into a typed Settings struct. We could follow the same pattern for the EWDS configuration instead of passing environment-variable names and defaults through every `query_via_ewds` call

offchain-primitives/src/ewds.rs
```diff
@@ -0,0 +1,373 @@
...
+            }),
+            attributes: order.attributes.map(|attributes| EwdsAttributesDto {
+                trading_partner_id: attributes.trading_partner_id,
+                energy_type: energy_type_to_ewds(&attributes.energy_type).to_string(),
+            }),
+        }
+    }
+}
+
+pub async fn query_via_ewds<T: DeserializeOwned>(request: EwdsQueryRequest) -> Result<Vec<T>> {
```
spyrostz: IMO this method is too dense and contains multiple concerns, for example it is: 1) parsing the environment variables 2) setting up the connection 3) sending the query to EWDS 4) polling for the response. IMO these 4 is better to be split into separate methods of a single struct, for readability.

offchain-primitives/src/ewds.rs
```diff
@@ -0,0 +1,373 @@
...
+    value.extend(suffix.chars().filter(|ch| ch.is_ascii_alphanumeric()));
+
+    if value.is_empty() {
+        base.to_string()
+    } else {
+        value
+    }
+}
+
+fn order_type_to_ewds(order_type: &OrderEnum) -> &'static str {
```
spyrostz: Also something else relevant to code organization, IMO the DTO types and the conversion functions can be in a separate module, and this module could only host the struct that manages the EWDS connection / messaging.

gsy-execution-engine/src/connectors/offchain_storage.rs
spyrostz: Generally I agree with adding a configuration layer in the execution engine too. However, I am somehow skeptical about the request / response topics being held in this variable. The query type is already defined by `operation` member, therefore there is no need to add the request / response topics as  EWDSQueryRequest members. A simple mapping function that would map the operation to a request / response topic would be sufficient and would not clutter that much the EWDSQueryRequest constructor.

spyrostz (commented):
Some refactoring changes and merge conflicts pending, otherwise it is ready to merge. Will LGTM once resolved

offchain-primitives/src/ewds/mod.rs
hannesdiedrich, line 1: NIT: Don't we want to group all ewds related code? E.g. this module could live as `mod.rs` in the ewds folder. Or is this not a good pattern?

offchain-primitives/src/ewds.rs
```diff
@@ -39,277 +54,296 @@ impl fmt::Display for EwdsOperation {
...
-    #[serde(default)]
-    pub attributes: Option<EwdsAttributesDto>,
+#[derive(Debug, Clone)]
+pub struct EwdsClientConfig {
+    pub gateway_base: String,
+    pub request_fqcn: String,
+    pub response_fqcn: String,
+    pub topic_owner: String,
+    pub topic_version: String,
+    pub response_client_id: String,
```
hannesdiedrich: This is a question I always had: Why is this called the `response_client_id` ? AFAIU, we are sending the name of the requesting service here, don't we? So I would expect something like `request_client_id` or similar. What do you think? (nothing urgent to be changed now if you agree).

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -125,32 +125,19 @@ async fn fetch_trades_and_measurements_via_ewds(
...
     });
+    let ewds_client = EwdsClient::from_env(
+        "EWDS_EXECUTION_ENGINE_CLIENT_ID",
+        "gsyexecutionengine",
+        8_000,
+    );
+
+    let trades: Vec<TradeSchema> = ewds_client
+        .query(EwdsOperation::TradesQuery, query.clone())
+        .await?;
```
hannesdiedrich, line 136: Beautiful! Thank you!

offchain-primitives/src/ewds/mod.rs
keyanhov: Agreed. Both layouts are valid Rust patterns, but now that EWDS contains its own DTO submodule, grouping the implementation under ewds/mod.rs is clearer. I'll move it there, the public module API will remains unchanged

offchain-primitives/src/ewds.rs
keyanhov: The value identifies the requesting service, but DDHub uses it on GET /api/v2/messages as the consumer/cursor ID while polling the response subscription, it is not included in the published request. I agree `request_client_id` could be misleading. I'll rename the field to `consumer_client_id`

keyanhov: addressed in [967d9c3](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/58/commits/967d9c3099edd7baff5f8f1e650f05256d716ca8)

offchain-primitives/src/ewds/mod.rs
keyanhov: addressed in [967d9c3](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/58/commits/967d9c3099edd7baff5f8f1e650f05256d716ca8)

PR #59, opened 2026-07-03
DD-404: Fix e2e tests

hannesdiedrich:


gsy-offchain-storage/src/http_server.rs
```diff
@@ -39,6 +38,7 @@ pub fn run(
             .route("/trades", web::post().to(post_trades))
             .route("/trades", web::get().to(get_trades))
             .route("/market", web::get().to(get_market))
+            .route("/markets", web::post().to(post_market))
```
hannesdiedrich, line 44: Why did you re-add this endpoint again?
The market orchestrator will create markets. This will be handles in the frame of DD-398.

keyanhov: the E2E flow started calling `AreaMarketInfoAdapter::create_market(...)`, which builds a `MarketSchema` and posts it to `/markets`. At that point the off-chain storage exposed only GET `/markets`, so the request returned 404, `create_market()` returned `None`, and the scenario failed

hannesdiedrich: thanks a lot for explaining!

hannesdiedrich (commented):
LGTM from my side

PR #70, opened 2026-07-28
Feat/dd 408

keyanhov:


primitives/src/ewds/mod.rs
```diff
@@ -245,15 +248,39 @@ impl EwdsClient {
...
-            .json(&send_message_body)
-            .send()
-            .await?;
-        let send_status = send_response.status();
-        if !send_status.is_success() {
+        let mut rate_limit_attempt = 0u32;
+        loop {
+            if started.elapsed() > Duration::from_millis(self.config.timeout_ms) {
+                return Err(anyhow!(
+                    "EWDS timeout sending {} request (request_id={})",
```
hannesdiedrich, line 255: Just out of curiosity: Did you actually experience this or is this just a safety net?

keyanhov: yes, in some of the e2e test run i experienced this

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -114,38 +122,54 @@ pub async fn start_ewds_request_handler(db: DatabaseWrapper, config: EwdsHandler
...
-        }
+            Ok(()) => {
+                rate_limit_attempt = 0;
+                config.poll_interval_ms
+            }
+            Err(error) => {
+                warn!("EWDS batch processing failed: {}", error);
+                if is_rate_limited_message(error.to_string().as_str()) {
+                    let delay_ms = ewds_rate_limit_backoff_ms(rate_limit_attempt);
+                    rate_limit_attempt = rate_limit_attempt.saturating_add(1);
```
hannesdiedrich, line 148: All good for now, but shouldn't we also cap the number of retries? I understand that `rate_limit_attempt` is used to calculate the `ewds_rate_limit_backoff_ms` and the latter is capped for 30000 ms. But it will never stop trying here. Should we maybe panic in case of 1000 retries (rate_limit_attempt>1000)?
If you agree I would put all this in a utility method called `get_poll_intervall` . 
But again, does not need to be fixed now, could be done i a separate ticket.

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -114,38 +122,54 @@ pub async fn start_ewds_request_handler(db: DatabaseWrapper, config: EwdsHandler
...
 ) -> Result<()> {
     let amount = config.request_batch_size.to_string();
-    let mut messages = Vec::new();
-    for operation in EwdsOperation::ALL {
-        let topic_name = config.topics.for_operation(operation).request.as_str();
-        messages
-            .extend(poll_requests_for_topic(client, config, topic_name, amount.as_str()).await?);
-    }
+    let operation = EwdsOperation::ALL[*next_topic_index % EwdsOperation::ALL.len()];
+    *next_topic_index = (*next_topic_index + 1) % EwdsOperation::ALL.len();
```
hannesdiedrich, line 170: Why changing this per reference and not looping in the calling method? IMO this would be clearer. But I guess that you have a reason for this. Looking forward to the answer.

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -421,10 +450,39 @@ async fn send_message_with_fqcn(
...
         let error_body = response.text().await.unwrap_or_default();
+        if is_rate_limited_response(status, error_body.as_str()) {
+            let delay_ms = ewds_rate_limit_backoff_ms(rate_limit_attempt);
+            warn!(
+                "EWDS rate limit while publishing response for topic '{}'; retrying in {} ms",
+                request_body.topic_name, delay_ms
+            );
+            rate_limit_attempt = rate_limit_attempt.saturating_add(1);
+            sleep(Duration::from_millis(delay_ms)).await;
+            continue;
```
hannesdiedrich, line 487: I think my proposal from above makes sense now, AFAIS, this part could live in a utility method that could be re-used also further up in this module. Or am I missing something?

docs/platform/execution-engine.md
```diff
@@ -32,6 +32,14 @@ Before submitting, the engine checks on-chain `penaltyEnergyByTrade(tradeId)`:
 
 This prevents repeated submission across recurring execution cycles.
 
+## Timeslot Rollover
+
+An EWDS request can remain in flight while the configured target delivery slot
+advances. The engine retains the outgoing target slot for a bounded number of
+additional cycles so trades indexed around the boundary are not skipped. It
+stops retrying as soon as penalties for that slot are submitted or found
+on-chain.
```
hannesdiedrich, line 41: I have the feeling that we are fixing things that are only due to the large RTT of the EWDS. 
Could we skip all this if the EWDS would be a true message broker with return times lower than 1 second?

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: You’re right. The two call sites have different retry ownership, so I would keep the logging and sleeping in their respective flows, but the backoff calculation plus attempt increment is duplicated. I’ll extract that into a small shared helper such as next_rate_limit_backoff_ms(&mut attempt) and reuse it here and in the polling loop.

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: The reason for processing one topic at a time was to avoid sending three consecutive gateway requests on every polling interval, which was contributing to the EWDS 429 responses. The index is passed by reference only to preserve round-robin state between iterations.
That said, I agree this scheduling responsibility belongs in the caller. I can make process_topic_batch accept an EwdsOperation and let the outer handler loop iterate over EwdsOperation::ALL, sleeping after each operation. This keeps the requests paced while making the control flow clearer and removing the mutable index parameter

gsy-execution-engine/src/main.rs
```diff
@@ -70,7 +163,7 @@ async fn main() {
     }
 }
 
-fn generate_previous_timeslot(_market_duration: u64) -> u64 {
+fn generate_target_timeslot() -> u64 {
```
hannesdiedrich: ❤️

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: I agree that the polling/backoff policy could be encapsulated in a utility such as get_poll_interval or a small backoff state object. However, I would not panic after an arbitrary number of retries. This is a long-running consumer and EWDS throttling may recover without intervention. Restarting the container would reset the capped backoff and could make the request burst worse. If we want fail-fast behavior, the handler should return an error cleanly rather than panic. I agree this can be handled in a separate ticket

docs/platform/execution-engine.md
keyanhov: Yes, the high EWDS RTT is what exposed this race in the E2E flow. If EWDS guaranteed that requests, responses, and resulting indexed data were visible before the slot boundary, the rollover retry would not be necessary. My concern is that low broker RTT alone does not provide that guarantee. Matching, on-chain event propagation, indexing, or consumer lag can still make a trade visible shortly after the target slot advances. The current protection is bounded to two retries, idempotent through the on-chain penalty check, and can be disabled with EXECUTION_ENGINE_ROLLOVER_RETRY_LIMIT=0. I would update the documentation to describe this as protection against asynchronous indexing around a slot boundary, rather than specifically an EWDS latency workaround. If EWDS provides a stronger completion/visibility guarantee, we can safely remove it

gsy-matching-engine/src/algorithms/pay_as_bid.rs
```diff
@@ -1,15 +1,38 @@
-use super::PayAsBid;
+use super::{ClearingPoint, PayAsBid};
 use crate::models::{BidOfferMatch, MatchingData, Order};
 use rand::Rng;
 use std::collections::HashMap;
 
 impl MatchingData {
-    fn match_preferences(&self) -> (Vec<BidOfferMatch>, Vec<Order>, Vec<Order>) {
+    pub(super) fn orders_for_market_slot(&self) -> (Vec<Order>, Vec<Order>) {
```
hannesdiedrich: I understand that the `pub(super)` makes the method accessible for the parent module. For me (coming from python) this is a weird concept. I would put these methods in the parent module and re-use them in the children. Is this not a good pattern in rust?

hannesdiedrich: Also another NIT: I would call this method `get_orders_for_market_slot` or `fetch_order_for_market_slot` in order to show directly on the calling side that this includes a request to the offchain storage.

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich, line 110: I have a general architecture question here:
Why aren't we fetching the orders for a specific market slot here? Wouldn't it reduce the amount of data sent in one batch? Or is this a non-issue and we would like to always match open orders?
I just think about a corner case: What if orders that have not been matched were not changed from "open" to another status (BTW; which process does that?) ? Wouldn't the number of markets returned here increased steadily ?

gsy-matching-engine/src/connectors/evm_connector.rs
```diff
@@ -492,6 +582,20 @@ fn to_evm_matches(
...
+                || item.time_slot != item.bid.time_slot
+            {
+                return Err(anyhow!(
+                    "Refusing cross-market/time-slot match between bid {}/{} and offer {}/{}",
+                    item.bid.market_id,
+                    item.bid.time_slot,
+                    item.offer.market_id,
+                    item.offer.time_slot
+                ));
+            }
```
hannesdiedrich, line 549: How can this happen? Doesn't `partition_orders_by_market_slot` deal with this?

gsy-matching-engine/src/models.rs
```diff
@@ -46,4 +46,28 @@ pub struct MatchingData {
     pub bids: Vec<Order>,
     pub offers: Vec<Order>,
     pub market_id: String,
+    pub time_slot: u64,
+}
+
+impl MatchingData {
+    pub fn validate_market_slot(&self) -> Result<(), String> {
```
hannesdiedrich: Is there a way of calling this on insert in order to not having to depend on calling this method?

gsy-matching-engine/src/algorithms/pay_as_bid.rs
```diff
@@ -124,10 +165,26 @@ impl MatchingData {
...
         for offer in &offers {
             for bid in &bids {
-                if offer.area_uuid == bid.area_uuid || offer.energy_rate > bid.energy_rate {
+                if remaining_clearing_energy == 0 {
+                    return matches;
+                }
+
+                if offer.area_uuid == bid.area_uuid
+                    || offer.market_id != bid.market_id
+                    || offer.time_slot != bid.time_slot
```
hannesdiedrich: These two checks should also not be needed at this point right?

docs/platform/matching-engine.md
```diff
@@ -9,25 +9,55 @@ to `TradeSettlement`.
...
+   with preference phase first.
+5. Build EVM tuple payload for `settleBatch`.
+6. Submit transaction with matching engine signer.
+
+## Matching Algorithms
+
+`MATCHING_ALGORITHM` selects the standard-market algorithm:
+
+- `pay_as_bid` (default): each accepted standard match settles at the bid rate.
+- `pay_as_clear`: bids are sorted descending and offers ascending, then the
```
hannesdiedrich:
```suggestion
- `pay_as_clear`: bids are sorted by descending energy_rate and offers by ascending energy_rate, then the
```

primitives/src/lib.rs
```diff
@@ -37,6 +39,44 @@ pub enum MatchingAlgorithm {
...
+            "pay_as_clear" | "pay-as-clear" => Ok(Self::PayAsClear),
+            "amm" => Ok(Self::AMM),
+            _ => Err(format!(
+                "Unsupported matching algorithm '{}'. Expected pay_as_bid, pay_as_clear, or amm",
+                value
+            )),
+        }
+    }
+}
+
```
hannesdiedrich, line 80: Do these need to live here in the primitives? I understand that we only need them in the e2e test (besides the matching engine itself)

e2e-tests/src/steps/trade_steps.rs
```diff
@@ -77,53 +82,79 @@ abigen!(
...
+            .request::<_, U256>("evm_mine", None::<()>)
             .await
-            .expect("Failed to await synthetic activity receipt");
+            .expect("Failed to mine an empty Anvil block");
     }
 }
 
-async fn emit_until_matching_block(world: &MyWorld, max_blocks: usize) {
+async fn mine_until_matching_block(world: &MyWorld, max_blocks: usize) {
+    let matching_block_interval = matching_engine_block_interval();
```
hannesdiedrich: why not re-using the one in matching_engine/evm_connector?

e2e-tests/src/steps/trade_steps.rs
```diff
@@ -77,53 +82,79 @@ abigen!(
...
     }
 
     panic!(
-        "Could not reach a matching trigger block after emitting {} synthetic blocks",
+        "Could not reach a matching trigger block after mining {} blocks",
         max_blocks
     );
 }
 
+fn matching_engine_block_interval() -> u64 {
```
hannesdiedrich: why not re-using the one in matching_engine/evm_connector?

e2e-tests/src/steps/trade_steps.rs
```diff
@@ -431,7 +462,88 @@ async fn submit_cheaper_offer(world: &mut MyWorld, user_name: String, energy: f6
...
 
     // Trigger matching after all preference/open-market orders were submitted.
-    emit_until_matching_block(world, 12).await;
+    mine_until_matching_block(world, 12).await;
+}
+
+#[when("the pay-as-clear order book is submitted")]
+async fn submit_pay_as_clear_order_book(world: &mut MyWorld) {
+    // A uniform-price auction must observe the complete book in one interval.
+    ensure_matching_window_capacity(world, 8).await;
```
hannesdiedrich: If this method panics, the step fails and the e2e test, too, right? Shouldn't we just wait for the sufficient block? Or did I get it wrong?

docs/setup/test.md
```diff
@@ -33,6 +33,37 @@ docker compose --env-file contracts-output/addresses.env \
...
+price, and both orders beyond that crossing remain open. The combined scenario
+also verifies that a preferred bilateral trade keeps its negotiated rate while
+the remaining standard order book clears at one uniform price. Both scenarios
+verify on-chain settlement and non-zero execution-engine penalties for every
+accepted trade.
+`pay_as_clear` defaults to a `64`-block matching interval so all scenario
+orders are collected before one clearing cycle. Override it with a positive
+`MATCHING_ENGINE_BLOCK_INTERVAL` value only when both the matching engine and
+E2E runner use the same value, as they do through the Compose configuration.
+After the complete order book is indexed, the E2E harness fast-forwards the
```
hannesdiedrich: >the E2E harness fast-forwards the
local Anvil chain with empty-block RPC calls rather than waiting for one
transaction per block.

Can you please point me to this in the code?

hannesdiedrich (commented):
Thanks a lot for this complex PR (for me).
Left some questions. Will review again after Spyros also had a look.

gsy-matching-engine/src/algorithms/pay_as_bid.rs
keyanhov: pub(super) is idiomatic Rust and intentionally limits visibility to the parent module, but I agree with the organizational point here. These helpers are shared by pay-as-bid and pay-as-clear and are not specifically pay-as-bid behavior. Moving the common matching logic into the parent algorithm module, or a dedicated common module, would make that clearer. orders_for_market_slot may become unnecessary entirely once MatchingData is validated when constructed

keyanhov: > Also another NIT: I would call this method get_orders_for_market_slot or fetch_order_for_market_slot in order to show directly on the calling side that this includes a request to the offchain storage.

This method does not perform an off-chain-storage request. The request happens earlier in run_matching_cycle; this method only filters the already-loaded bids and offers vectors. Therefore, get_ or fetch_ would suggest I/O that does not happen. If we retain it, filter_orders_for_market_slot would be more explicit, although validated construction should make the filtering unnecessary

gsy-matching-engine/src/connectors/evm_connector.rs
keyanhov: The existing order API already supports market and time filters, but it does not yet support a status filter and the matching engine currently sends no filters. I suggest handling that as a focused follow-up rather than mixing the query/lifecycle change into the pay-as-clear implementation. IMO, we should create a specific ticket for this

gsy-matching-engine/src/connectors/evm_connector.rs
keyanhov: In the current execution path, partition_orders_by_market_slot and validate_market_slot should make this mismatch impossible. The check here is deliberately at the final contract boundary, so an algorithm regression or another future caller cannot submit a cross-market match on-chain. I agree we do not need the same invariant checked at every layer. Once MatchingData is validated by construction, I would remove the duplicate algorithm-level checks but retain one final boundary check before settleBatch

gsy-matching-engine/src/models.rs
keyanhov: Yes. Because the vectors and fields are currently public, Rust cannot automatically validate direct construction or Vec::push. The idiomatic approach is to make the fields private and provide MatchingData::new(...) -> Result<Self, _>, plus validated add_bid/add_offer methods if mutation is needed. That would make invalid market/time-slot combinations unrepresentable through the public API and remove the need to remember an explicit validation call

gsy-matching-engine/src/algorithms/pay_as_bid.rs
keyanhov: Correct for the current connector path, after partitioning and validation, the market and time-slot checks are redundant. They only protect us because MatchingData can currently be constructed directly with public fields. After introducing validated construction, these checks can be removed while retaining the area and price compatibility checks

primitives/src/lib.rs
keyanhov: This type is not only used by the matching engine and E2E tests. It is part of the shared serialized MarketSchema, and the community client supplies it when creating a market. Therefore, keeping the enum and its serialized representation in primitives is appropriate, only the actual algorithm implementation belongs in the matching-engine crate

e2e-tests/src/steps/trade_steps.rs
keyanhov: The duplication is worth removing, but the current matching-engine function is private inside evm_connector, and the E2E crate does not depend on the matching-engine crate. Adding the entire service as an E2E dependency only for this helper would introduce unnecessary compile-time coupling. I suggest extracting the interval resolution into a small shared configuration helper so both processes read the same policy without importing the connector

e2e-tests/src/steps/trade_steps.rs
keyanhov: Yes, the panic fails the Cucumber step. It was intentionally fail-fast because crossing a matching boundary while submitting the book would split a uniform-price auction across two cycles. However, failing is unnecessarily brittle. A better helper would detect insufficient remaining capacity, explicitly advance or wait until the next matching boundary, and then submit the complete book at the beginning of the new interval

docs/setup/test.md
keyanhov: The empty-block fast-forwarding is implemented in [mine_empty_blocks](https://github.com/gridsingularity/gsy-decentralized-exchange/blob/feat/DD-408/e2e-tests/src/steps/trade_steps.rs#L85), which calls Anvil’s evm_mine RPC. mine_until_matching_block repeatedly uses it until the configured matching boundary is reached. It is then called after the scenario orders have been indexed

gsy-matching-engine/src/algorithms/pay_as_bid.rs
hannesdiedrich: Ah yes, you are right.
> filter_orders_for_market_slot
makes more sense. Thanks!

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich: Agreed, this should be done in a separate ticket.

primitives/src/lib.rs
hannesdiedrich: OK, understood, Thanks!

e2e-tests/src/steps/trade_steps.rs
hannesdiedrich: So, what is your plan here? Should we restart the e2e tests in this case or are you changing this? (Sorry, did not get this from "A better helper would detect insufficient remaining capacity...")

keyanhov: No restart is needed. I’m changing this so when there is insufficient capacity left in the current interval, the E2E runner advances Anvil to the next matching boundary before submitting any orders, then continues the same test run. It only fails if the configured interval is fundamentally too small to contain the order submission window

gsy-offchain-storage/src/ewds_handler.rs
hannesdiedrich: Leaving this open for @spyrostz to comment

gsy-offchain-storage/src/ewds_handler.rs
hannesdiedrich: I do not see this implemented. Did you decide against it?

docs/platform/execution-engine.md
hannesdiedrich: Also leaving this open for Spyros to read.

gsy-matching-engine/src/algorithms/mod.rs
```diff
@@ -43,3 +39,210 @@ struct ClearingPoint {
...
+
+    let mut residual = order.clone();
+    residual.energy -= matched_energy;
+    if matched_energy > 0 {
+        residual.order_id = random_bytes16_hex();
+    }
+    Some(residual)
+}
+
+fn random_bytes16_hex() -> String {
```
hannesdiedrich, line 246: I only see this now and I understand that this is heritage code
IMO this method should reside in the primitives::utils and we should decide whether we use this random approach or the one that I did [in the other PR](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/74/changes#diff-7fc4fc81cf1fa6c254ffc51687cbac9600a6fb73023db96826e21f9b2801b3daR140-R141) using UUID.
In order to have one way of creating random ids. 
Just mentioning it here in order to not forget before merging (whoever merges first).

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich: Also Spyros to decide. Happy to open a ticket if agreed.

gsy-matching-engine/src/models.rs
```diff
@@ -43,31 +43,86 @@ pub struct BidOfferMatch {
...
+    offers: Vec<Order>,
+    market_id: String,
+    time_slot: u64,
 }
 
 impl MatchingData {
-    pub fn validate_market_slot(&self) -> Result<(), String> {
-        let mismatched_order = self
-            .bids
+    pub fn new(
```
hannesdiedrich, line 53: ❤️

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich: OK, the latter makes sense. Then let's leave this here.

e2e-tests/src/steps/trade_steps.rs
hannesdiedrich: Ah, now I understand. thanks!

e2e-tests/features/pay_as_clear/pay_as_clear_matching.feature
```diff
@@ -0,0 +1,31 @@
...
+    And the matching engine uses "pay_as_clear"
+    And users "alice", "bob", and "charlie" the matching engine operator are registered
+    When the Market Orchestrator opens the Spot market for the next delivery slot
+    And the community market and forecasts of 10 energy are submitted
+    And measurements for facilities are submitted
+    And the pay-as-clear order book is submitted
+    Then the market clears 7 energy at a uniform price of 10
+    And the matching engine matches the bid and offer and a trade is settled on-chain
+    And orders beyond the clearing point remain open
+    And the execution engine submits penalties for the trade
```
spyrostz, line 17: I am missing a step for the transition of the Trade status from "Settled" to "Executed". This can be done in a separate ticket though if the EVM handler does not manage the state transitions of the Trade object, let me know to create one in this case!

gsy-execution-engine/src/main.rs
```diff
@@ -9,6 +9,84 @@ use gsy_execution_engine::{
...
+const DEFAULT_ROLLOVER_RETRY_LIMIT: u32 = 2;
+
+#[derive(Debug)]
+struct PendingTimeslot {
+    timeslot: u64,
+    retries_remaining: u32,
+}
+
+#[derive(Debug)]
+struct ExecutionTimeslotScheduler {
```
spyrostz: I think that it is better for readability for the timeslot scheduler and its related methods to live in a separate module.

gsy-execution-engine/src/main.rs
```diff
@@ -43,14 +121,23 @@ async fn main() {
...
+                .and_then(|value| value.parse::<u32>().ok())
+                .unwrap_or(DEFAULT_ROLLOVER_RETRY_LIMIT);
+            let initial_timeslot = generate_target_timeslot();
+            let mut timeslot_scheduler =
+                ExecutionTimeslotScheduler::new(initial_timeslot, rollover_retry_limit);
+
             loop {
-                let timeslot = generate_previous_timeslot(market_duration);
+                let current_target_timeslot = generate_target_timeslot();
+                let timeslot = timeslot_scheduler.next_timeslot(current_target_timeslot);
```
spyrostz: Same as above, I think that it is better to have a single struct called TimeslotScheduler or similar, with a calculate_timeslot() method that will calculate and return only the timeslot. That way the main function will remain clean.

gsy-execution-engine/src/main.rs
```diff
@@ -79,3 +172,42 @@ fn generate_previous_timeslot(_market_duration: u64) -> u64 {
 
     (prev.timestamp() as u64 / GLOBAL_CONSTANTS.time_slot_sec) * GLOBAL_CONSTANTS.time_slot_sec
 }
+
+#[cfg(test)]
+mod tests {
```
spyrostz: General comment, we do not have a strict rule but I think that since most of the other tests are in separate modules and directories, the same should apply for these, and they should be moved under the tests/ dir.

gsy-matching-engine/src/algorithms/mod.rs
```diff
@@ -1,7 +1,248 @@
...
+                    continue;
+                }
+
+                matches.push(BidOfferMatch {
+                    market_id: offer.market_id.clone(),
+                    time_slot: offer.time_slot,
+                    bid: bid.clone(),
+                    offer: offer.clone(),
+                    residual_bid: None,
+                    residual_offer: None,
```
spyrostz, line 104: I assume that here the residual orders are None because of the preference matching, and because they could be matched also in the normal matching later, correct?

gsy-matching-engine/src/algorithms/pay_as_clear.rs
```diff
@@ -0,0 +1,205 @@
...
+
+            let bid = bids[bid_index];
+            let offer = offers[offer_index];
+            if bid.energy_rate < offer.energy_rate {
+                break;
+            }
+
+            let accepted_energy = bid_energy.min(offer_energy);
+            traded_energy += accepted_energy;
+            clearing_price = Some(offer.energy_rate);
```
spyrostz, line 45: In general the algorithm is okay, however it is not handling 2 corner cases, where the energy of the supply or demand is exhausted before the clearing point is reached. Most pay as clear algorithms usually use either the offer, bid or midpoint price when a clearing point is reached, however in case of scarcity of supply or demand, they use the minimum bid price and maximum offer price respectively. 
Having said that, I recommend to deal with this in a separate ticket, where we will do more elaborate configuration of the pay as clear matching algorithm. Created a ticket here https://gridsingularity.atlassian.net/browse/DD-422

gsy-matching-engine/src/connectors/evm_connector.rs
spyrostz: I agree that this should be a separate ticket https://gridsingularity.atlassian.net/browse/DD-423

spyrostz (commented):
Left some comments, will LGTM once resolved

gsy-matching-engine/src/algorithms/mod.rs
hannesdiedrich: We have declared as a bug actually that is planned to be fixed here
https://gridsingularity.atlassian.net/browse/DD-420

e2e-tests/features/pay_as_clear/pay_as_clear_matching.feature
keyanhov: we're currently missing the 'Settled' status in the contract, I'll need to update both contract and handler. Let's create a specific ticket for this

hannesdiedrich: There is a conversion function between the ontology (that supports "settled") and the contract. 
This specific conversion will change in the DD-425 BTW. So I do not think that we need to support "settled" in the smart contract but use "executed". 
What do you all think?

spyrostz (approved):
LGTM

PR #74, opened 2026-08-07
Feat/dd 410

hannesdiedrich:


primitives/src/ewds/mod.rs
```diff
@@ -82,6 +88,10 @@ impl Default for EwdsTopicConfig {
                 request: "measurementsQuery".to_string(),
                 response: "measurementsQueryResponse".to_string(),
             },
+            ids: EwdsTopicPair {
+                request: "idsQuery".to_string(),
```
hannesdiedrich, line 92: TODO for myself: these have to be added to the wiki on confluence for the deploy to the staging env.

keyanhov:
We should also expose the ID mapping operation over HTTP, alongside ids.query, to remain consistent with the existing OFFCHAIN_STORAGE_TRANSPORT abstraction. HTTP deployments should be able to get/create and retrieve mappings without requiring EWDS configuration

gsy-offchain-storage/src/db/id_service.rs
```diff
@@ -0,0 +1,99 @@
...
+use mongodb::{Collection, IndexModel};
+use primitives::db_api_schema::ids::IdMappingSchema;
+use primitives::utils::{bytes16_to_hex, create_encrypted_bytes16_from_string};
+use std::ops::Deref;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+pub async fn init_id_mapping(db: &DatabaseWrapper) -> Result<()> {
+    let controller = db.id_mapping();
+    controller
+        .create_index(IndexModel::builder().keys(doc! {"onchain_id": 1}).build())
```
keyanhov: These indexes should be unique. Without a unique offchain_id index, two concurrent find_one_and_update(...).upsert(true) calls can both insert a mapping. onchain_id should also be unique so a truncated-hash collision is detected rather than silently creating an ambiguous reverse mapping

gsy-community-client/src/node_connector/orders.rs
```diff
@@ -123,68 +130,62 @@ abigen!(
...
         (forecast.energy_kwh.abs() * rate_multiplier * NODE_FLOAT_SCALING_FACTOR) as u64,
         energy_type_to_contract(&EnergyType::None),
         energy_type_to_contract(&EnergyType::None),
         is_bid,
-    )
+    ))
 }
 
-pub fn create_input_orders(
+pub async fn create_input_orders(
```
keyanhov, line 161: This signature change breaks the existing node_orders tests because they call create_input_orders synchronously and immediately use .len() and indexing. At the moment the implementation performs no async operation because it hashes IDs locally. If the intended fix is to query ids.query, I'd suggest to add that call and convert these tests to async otherwise the function should remain synchronous

gsy-offchain-storage/tests/api/ids.rs
keyanhov, line 1: These tests validate the database service, but not the EWDS endpoint required by the ticket. Imo we should add ids.query success, malformed-payload and repeated-request cases to ewds_handler.rs tests, asserting the response topic, envelope and returned mapping

gsy-matching-engine/src/connectors/evm_connector.rs
```diff
@@ -109,9 +106,9 @@ async fn run_matching_cycle(
     let prepared_orders =
         fetch_open_orders_from_orderbook_service(orderbook_url.to_string()).await?;
     info!(
```
keyanhov, line 109: This print every order and match at INFO level. That is unrelated to DD-410, can expose actor/order data, and will increase log volume with large books. Can we retain the count-based logging or move detailed values to debug?

gsy-offchain-storage/src/db/id_service.rs
hannesdiedrich: I agree, that the `onchain_id` should be unique. Due to the current implementation, the `offchain_id` is then unique implicitly (because only the `get_or_create` deals with the hashing) . If you want to add a unique constraint for the `offchain_id` too, we need to think what happens if there is indeed a hash collision in the endpoint. 
Something worth mentioning here:
I remember that we decided to add the `creation_time` to the `IdMappingSchema` in order to allow versioning. But I can not remember how this should work @spyrostz , do you?

gsy-offchain-storage/tests/api/ids.rs
hannesdiedrich: Indeed, you are right. This PR was created before we started with adding the ewds_handler tests. Will add it here.

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich: Good catch!
I agree, this is a remnant of my tests. Will revert it.

hannesdiedrich:
> We should also expose the ID mapping operation over HTTP, alongside ids.query, to remain consistent with the existing OFFCHAIN_STORAGE_TRANSPORT abstraction. HTTP deployments should be able to get/create and retrieve mappings without requiring EWDS configuration

I have added an http endpoint [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/74/commits/5798e3ddd7d3e981c55a9d543792f95250ca97be)

gsy-offchain-storage/src/db/id_service.rs
hannesdiedrich: I made the indexes unique here https://github.com/gridsingularity/gsy-decentralized-exchange/pull/74/commits/5798e3ddd7d3e981c55a9d543792f95250ca97be

gsy-matching-engine/src/connectors/evm_connector.rs
hannesdiedrich: Done [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/74/commits/1ceb5b2feb4b78651d9dedf13503aa935ad1fee5)

gsy-community-client/src/node_connector/orders.rs
hannesdiedrich: Done [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/74/changes/f1014df10011ad6deb5491a3e655d546d08e215e)

gsy-offchain-storage/src/db/id_service.rs
keyanhov: lgtm, i'll leave the comment for @spyrostz to answer your previous question

keyanhov (approved):
LGTM

gsy-community-client/tests/evm_integration.rs
```diff
@@ -172,6 +172,7 @@ async fn test_publish_orders_calls_evm_order_registry() {
         compile_and_deploy_contract(client.clone(), source, "MockOrderRegistry").await;
 
     let market = test_market();
+    eprintln!("{:?}", market);
```
spyrostz: Reminder to remove before merge

gsy-execution-engine/src/connectors/evm_connector.rs
```diff
@@ -27,9 +26,10 @@ fn to_evm_penalties(penalties: Vec<Penalty>) -> Vec<EvmPenaltyTuple> {
             }
 
             Some((
-                parse_or_hash_bytes16(&penalty.penalized_account),
-                parse_or_hash_bytes16(&penalty.market_id),
-                parse_or_hash_bytes16(&penalty.trade_uuid),
+                parse_uuid_or_hex_bytes16(&penalty.penalized_account)
+                    .expect("failed to parse uuid"),
+                parse_uuid_or_hex_bytes16(&penalty.market_id).expect("failed to parse uuid"),
+                parse_uuid_or_hex_bytes16(&penalty.trade_uuid).expect("failed to parse uuid"),
```
spyrostz: Minor, but if you really expect this to fail under certain conditions, I would not use expect on this, but just skip the entry from the `collect`. If this is an extreme condition and it is considered a bug, then leave it as is.

gsy-community-client/src/node_connector/orders.rs
```diff
@@ -123,68 +130,62 @@ abigen!(
...
-    let order_id = parse_or_hash_bytes16(
-        format!(
-            "{}:{}:{}:{}:{}",
-            market.market_id, facility_id, market.delivery_start_time, index, is_bid
-        )
-        .as_str(),
-    );
+    let offchain_order_id = Uuid::new_v4().to_string();
+    let onchain_order_id = create_encrypted_bytes16_from_string(&offchain_order_id);
+    let onchain_facility_id = create_encrypted_bytes16_from_string(facility_id);
```
spyrostz, line 143: This is actually what is puzzling me. I would expect the onchain order id and facility id to be saved in the offchain storage here, via an EWDS request or HTTP request to the /ids endpoint. I do not see this here though, can you please point me to the location if it is implemented already ? Thanks in advance

gsy-matching-engine/src/algorithms/mod.rs
```diff
@@ -238,11 +238,7 @@ fn residual_order(order: &Order, matched_energy: u64) -> Option<Order> {
     let mut residual = order.clone();
     residual.energy -= matched_energy;
     if matched_energy > 0 {
-        residual.order_id = random_bytes16_hex();
+        residual.order_id = Uuid::new_v4().to_string();
```
spyrostz, line 241: I think that this needs to be registered by explicitly calling the /ids endpoint or EWDS endpoint. Otherwise, the offchain storage will not be able to associate the offchain residual order id with the onchain one

spyrostz (commented):
Some comments, with 2 being really important. Will LGTM once they are clarified

gsy-community-client/src/node_connector/orders.rs
hannesdiedrich: We decided to only save actor ids in the DB, this is why I do not call the endpoints here.

gsy-execution-engine/src/connectors/evm_connector.rs
hannesdiedrich: I am not sure how to deal with this.
We could skip the penalties where these conversions fail with
```rust
fn to_evm_penalties(penalties: Vec<Penalty>) -> Vec<EvmPenaltyTuple> {
    penalties
        .into_iter()
        .filter_map(|penalty| {
            if penalty.penalty_cost == 0 {
                tracing::warn!(
                    "Skipping penalty for trade '{}' because penalty_cost is zero",
                    penalty.trade_uuid
                );
                return None;
            }

            Some((
                parse_uuid_or_hex_bytes16(&penalty.penalized_account).ok()?,
                parse_uuid_or_hex_bytes16(&penalty.market_id).ok()?,
                parse_uuid_or_hex_bytes16(&penalty.trade_uuid).ok()?,
                penalty.penalty_cost,
            ))
        })
        .collect()
}
```
What do you think?

gsy-community-client/src/node_connector/orders.rs
hannesdiedrich: Discussed offline, will mark as resolved.

gsy-offchain-storage/src/db/id_service.rs
spyrostz: The creation time is only useful in database migrations. Essentially, if we want to modify an onchain id, we could just add one row with the same offchain id, and the new onchain id. That way both ids would still be in the database, and we would be able to retrieve the valid onchain id even for times before the generation of the new onchain id based on the creation time of the id. This is premature optimization though, so we can leave the unique index and remove it if / when it becomes an actual requirement.

gsy-execution-engine/src/connectors/evm_connector.rs
spyrostz: Unfortunately this will not work if I see correctly. I meant to handle it the same way as the penalties, similar to this. It is a bit verbose but I think the logs will be helpful:

```rust
fn to_evm_penalties(penalties: Vec<Penalty>) -> Vec<EvmPenaltyTuple> {
    penalties
        .into_iter()
        .filter_map(|penalty| {
            if penalty.penalty_cost == 0 {
                tracing::warn!(
                    "Skipping penalty for trade '{}' because penalty_cost is zero",
                    penalty.trade_uuid
                );
                return None;
            }
             let Some(penalized_account) = parse_uuid_or_hex_bytes16(&penalty.penalized_account)
              else {
                  warn!(
                      "Skipping penalty for trade '{}': penalized_account '{}' is not a UUID",
                      penalty.trade_uuid, penalty.penalized_account
                  );
                  return None;
              };

              let Some(market_id) = parse_uuid_or_hex_bytes16(&penalty.market_id) else {
                  warn!(
                      "Skipping penalty for trade '{}': market_id '{}' is not a UUID",
                      penalty.trade_uuid, penalty.market_id
                  );
                  return None;
              };

              let Some(trade_uuid) = parse_uuid_or_hex_bytes16(&penalty.trade_uuid) else {
                  warn!(
                      "Skipping penalty for trade '{}': trade_uuid is not a UUID",
                      penalty.trade_uuid
                  );
                  return None;
              };

            Some((
                penalized_account,
                market_id,
                trade_uuid,
                penalty.penalty_cost,
            ))
        })
        .collect()
}
```

hannesdiedrich: OK, let me add this

spyrostz (approved):
LGTM

PR #78, opened 2026-08-19
Feat/dd 406

hannesdiedrich:


primitives/src/ewds/dto.rs
```diff
@@ -196,47 +185,125 @@ fn order_type_from_ewds(value: &str) -> Result<OrderEnum> {
...
-                energy_type: Some(IntelligentEnergyType::Green),
-                preferred_energy_rate: Some(11.0),
+                energy_type: Some(EnergyType::Green),
+                preferred_energy_rate: Some(12.0),
             }),
             attributes: None,
         }
     }
 
     #[test]
```
spyrostz: Minor and personal preference, but if you can move the tests to separate files would be great. I know that having the tests in the same file has the advantage that you are able to test private members, on the other hand though private members are not really needed to be tested :)

e2e-tests/src/steps/trade_steps.rs
```diff
@@ -627,7 +627,7 @@ async fn verify_charlie_offer_untouched(world: &mut MyWorld) {
         .expect("Charlie offer order was not found in off-chain storage");
     assert_eq!(
         charlie_offer.status,
-        OrderStatus::Open,
+        OrderStatus::PartiallyFilled,
```
spyrostz: This can also be postponed for later, but I think that the PartiallyFilled is actually a final state. Meaning that the new residual order should be marked as open, while the old order which has been partially matched should be marked as PartiallyFilled. Again, this remembering the discussion that we had with other partners about the matter.

primitives/src/ewds/dto.rs
```diff
@@ -306,6 +308,117 @@ impl TryFrom<EwdsTradeDto> for DbTradeSchema {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct EwdsClearingResultDto{
```
spyrostz: I assume that this is not populated in the DB ATM, correct ?

e2e-tests/src/steps/trade_steps.rs
hannesdiedrich: You are right. 
Later on in the process I changed it again
https://github.com/gridsingularity/gsy-decentralized-exchange/pull/78/changes#diff-39d63caa6380d111d95726ec23dfc9e5643bcfe5a85f5f8b32f6fac5518f01c3R638

primitives/src/ewds/dto.rs
hannesdiedrich: I don't know what you mean, but internally we use `ClearingResultSchema` which is also the DB schema (`ClearingResultService`)
We convert to and from `EwdsClearingResultDto` in the respective rest endpoints (e.g. [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/78/changes#diff-023120daad5b40748f8021ddcdf53a9d48a7838177c227d3c9851cd6130ae112R87))
I have started with DD-413 that deals with exposing a query for this to the EWDS too. 
I hope this answers your question

primitives/src/ewds/dto.rs
hannesdiedrich: OK, will do.

spyrostz (approved):
Minor comments, LGTM

primitives/src/ewds/dto.rs
spyrostz: Yes, did not mention that it is irrelevant to this ticket, just out of curiosity: is the `ClearingResultSchema` DB document populated by any service ? Because I do not remember having added such a functionality, and we should probably create a ticket to add it.

hannesdiedrich: Ah, now I see what you mean. AFAIS, No, not process writes to this document, not even in https://github.com/gridsingularity/gsy-decentralized-exchange/pull/70 . Let's create a ticket.

spyrostz (approved):
LGTM

PR #80, opened 2026-08-20
Feat/dd 411

keyanhov:


gsy-contracts/contracts/MarketController.sol
hannesdiedrich, line 33: @spyrostz When I first started with these changes, we decided to keep all information about the market onchain, so the `setMarketStatus` should actually be a `updateMarketInfo` and update a mapping between `marketId` and `MarketInfo` (the latter mirroring the information of the DB MarketSchema.
This is nothing for now of course, but should we add a ticket for this?

gsy-execution-engine/src/connectors/offchain_storage.rs
```diff
@@ -136,18 +136,15 @@ async fn fetch_trades_and_measurements_via_ewds(
         8_000,
     );
 
-    let trades: Vec<EwdsTradeDto> = ewds_client
-        .query(EwdsOperation::TradesQuery, query.clone())
-        .await?;
+    let (trades, measurements): (Vec<EwdsTradeDto>, Vec<MeasurementSchema>) = tokio::try_join!(
```
hannesdiedrich, line 139: Can you please explain the reason for `tokio::try_join!` here? 
These could also be two individual queries, right?

gsy-market-orchestrator/src/community_source.rs
```diff
@@ -0,0 +1,168 @@
...
+impl CommunityProvider for OffchainStorageCommunitySource {
+    async fn fetch_communities(&self) -> Result<Vec<EnergyCommunitySchema>> {
+        match self.transport {
+            OffchainStorageTransport::Http => self.fetch_via_http().await,
+            OffchainStorageTransport::Ewds => self.fetch_via_ewds().await,
+        }
+    }
+}
+
+#[cfg(test)]
```
hannesdiedrich: I think these tests will not be executed by the `gsy-market-orchestrator-integration-test` container. 
First, because of 
`entrypoint: ["cargo", "test", "--test", "evm_integration"]`
and second, because these tests are not in the test folder. 
We discussed this already in the past and I understand the reason for simply adding the tests in the module, however IMO these should live in the tests folder. We do not need specific test for non-public methods anyway.

gsy-market-orchestrator/src/community_source.rs
```diff
@@ -0,0 +1,168 @@
...
+    async fn fetch_communities(&self) -> Result<Vec<EnergyCommunitySchema>>;
+}
+
+pub struct OffchainStorageCommunitySource {
+    transport: OffchainStorageTransport,
+    offchain_storage_url: String,
+    http_client: reqwest::Client,
+}
+
+impl OffchainStorageCommunitySource {
```
hannesdiedrich, line 19: Can you please explain the reason for this rather complex implementation?
IN my other PR I added a method for fetching the facilities in a utility module [here](https://github.com/gridsingularity/gsy-decentralized-exchange/pull/83/changes#diff-a86e986a37ffbd28fc9d45da7dedd82466354b4b2952660dc5479d471f0b3905).
Is this not a good practice?
I see a lot of boilerplate code here, or is this the correct way to approach this in rust?

gsy-market-orchestrator/src/config.rs
```diff
@@ -16,6 +24,10 @@ pub struct Config {
     pub tick_interval_seconds: u64,
     #[serde(default = "default_look_ahead")]
     pub look_ahead_hours: u64,
+    #[serde(default)]
+    pub offchain_storage_transport: OffchainStorageTransport,
```
hannesdiedrich, line 28: I understand that the default is `Http`, right?
But how can I configure it to be `Ewds` or rather : Where is the `OFFCHAIN_STORAGE_TRANSPORT` env variable read here?

primitives/src/utils/mod.rs
```diff
@@ -74,3 +91,41 @@ pub fn string_to_timestamp(timestamp_string: &str) -> Result<u64> {
     let ts: u64 = timestamp_string.parse()?;
     Ok(ts)
 }
+
```
hannesdiedrich: I still vote for moving the tests in their own module.

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -372,6 +386,44 @@ pub async fn handle_request(
 
             send_success_response(client, config, request_id, response_topic.as_str(), data).await
         }
+        EwdsOperation::CommunityUpsert => {
```
hannesdiedrich, line 389: We should discuss at some point how we can reduce the number of lines added by each operation in order to not let the module grow too large. I foresee quite a lot of operations that we will add in the future.

gsy-offchain-storage/tests/api/ewds_handler.rs
```diff
@@ -20,7 +20,7 @@ fn test_config(gateway_url: String) -> EwdsHandlerConfig {
         topics: EwdsTopicConfig::from_env(),
         poll_interval_ms: 500,
         request_batch_size: 100,
-        response_send_timeout_ms: 60_000,
+        response_send_timeout_ms: 1_000,
```
hannesdiedrich, line 23: ❤️

schemas/ewds/intelligent/int.communities.query.request.v1.json
hannesdiedrich, line 1: I have never added or adapted these. Are you using AI to do so or is there a library that helps you?

AGENTS.md
```diff
@@ -1,3 +0,0 @@
-AI is not permitted in this repository.
```
hannesdiedrich, line 1: Did Claude remove this one? :-D
Or was it intenional?

CONTRIBUTING.md
```diff
@@ -32,10 +32,6 @@ Changes that are cosmetic in nature and do not add anything substantial to the s
 
 * If things are still unclear: Ask any question about how to use the GSY DEX in the [Grid Singularity Community Slack](https://gsycommunity.slack.com/join/shared_invite/zt-dpfxe2gk-hKFccsrr68g982odCJlrIg#/).
 
-#### **No AI**
-
-Do not contribute anything using generative AI tools.
```
hannesdiedrich, line 37: Also here?

e2e-tests/src/steps/market_steps.rs
```diff
@@ -147,3 +159,163 @@ async fn wait_for_market_to_open(world: &mut MyWorld) {
...
+
+    format!(
+        "{}-{}-{}-{}-{}",
+        &encoded[0..8],
+        &encoded[8..12],
+        &encoded[12..16],
+        &encoded[16..20],
+        &encoded[20..32]
+    )
+}
```
hannesdiedrich: Why not using Uuid also here?

e2e-tests/src/steps/trade_steps.rs
```diff
@@ -157,10 +161,11 @@ fn actor_id_as_hex(world: &MyWorld, user_name: &str) -> String {
...
 
 fn market_id_as_hex(world: &MyWorld) -> String {
-    format!(
-        "0x{}",
-        hex::encode(world.last_market_id.expect("Missing market id"))
-    )
+    market_id_bytes_as_hex(world.last_market_id.expect("Missing market id"))
+}
+
+fn market_id_bytes_as_hex(market_id: [u8; 16]) -> String {
```
hannesdiedrich: How about using the primitive utility `bytes16_to_hex` also here

gsy-contracts/contracts/MarketController.sol
hannesdiedrich: This will BTW solve the "issue":
In the evm_handler.rs in the gsy-offchain-storage we do not do anything than logging if a `MarketStatusUpdated` was received.

hannesdiedrich (commented):
I left some comments, will review again after answered/adapted.

I have a general comment, mainly for @spyrostz: We need to discuss/design and agree on the process which is creating entries in the `MarketSchema` document in the DB. Or am I missing that this is already done?

AGENTS.md
keyanhov: my bad, i inadvertently removed it

CONTRIBUTING.md
keyanhov: yes, my bad. Fixing in the coming commit

gsy-execution-engine/src/connectors/offchain_storage.rs
keyanhov: Yes, they could be two sequential queries. I used try_join! because the trades and measurements queries are independent, so there is no reason to wait for one before starting the other. This reduces the total EWDS round-trip/polling time

gsy-market-orchestrator/src/community_source.rs
keyanhov: You are right. I’ll move them to tests/community_source.rs and broaden the container command to run all integration-test targets, for example cargo test --tests

gsy-market-orchestrator/src/community_source.rs
keyanhov: The trait and source struct are mainly there to keep the orchestrator independent of HTTP/EWDS. The struct also keeps transport configuration explicit and reuses the HTTP client. The facility utility is not bad practice, it is a simpler pragmatic approach. My concern with using the same pattern here is that it reads environment variables and creates clients internally, which couples the caller to infrastructure and makes isolated orchestrator tests harder. I therefore prefer keeping the provider abstraction here, although I’m open to simplifying naming or boilerplate

gsy-market-orchestrator/src/config.rs
keyanhov: Correct, the default is http. The environment variable is read indirectly by `envy::from_env::<Config>()` in `get_config()`.  OFFCHAIN_STORAGE_TRANSPORT maps to the offchain_storage_transport field. Setting OFFCHAIN_STORAGE_TRANSPORT=ewds selects the Ewds variant

primitives/src/utils/mod.rs
keyanhov: I’ll move them to primitives/tests/utils.rs

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: Agreed. I suggest a to create a ticket/epic for this. We can move payload parsing and database work into operation-specific handlers/modules while keeping polling, deduplication, error handling, and response publishing shared

schemas/ewds/intelligent/int.communities.query.request.v1.json
keyanhov: I used the existing EWDS JSON schemas as templates and adapted them to the runtime DTOs. AI helped me with the initial drafting/checking. It would be useful to add automated JSON Schema validation in CI so correctness does not depend only on manual review

e2e-tests/src/steps/market_steps.rs
keyanhov: No strong reason. I’ll change it

e2e-tests/src/steps/trade_steps.rs
keyanhov: Agreed

gsy-execution-engine/src/connectors/offchain_storage.rs
hannesdiedrich: Great! Keeping this in mind for future implementations.

gsy-market-orchestrator/src/community_source.rs
hannesdiedrich: Sorry if I am asking weird questions, but my software development knowledge is limited here. I like to learn for the future.
> The struct also keeps transport configuration explicit 

What do you mean specifically? What would be a implicit configuration? AFAIU, we use the OFFCHAIN_STORAGE_TRANSPORT env variable for toggling between ewds and pure http .

> and reuses the HTTP client.

What do you mean by re-using? Do you mean that it is only initiated once when the whole program is started and we only call `reqwest::Client::new()` once?

> which couples the caller to infrastructure and makes isolated orchestrator tests harder. 

Please explain a bit more. Why is reading a env variable coupling to infrastructure?

gsy-market-orchestrator/src/config.rs
hannesdiedrich: Puh ok, this was not directly understandable to me. Not a fan, but as this is personal preference, I will resolve the comment.

schemas/ewds/intelligent/int.communities.query.request.v1.json
hannesdiedrich: So we never actually use these to check the received (and sent) DTOs, right?
Was this a requirement from the project or are these remnant of former versions of the repo?
IMO, if we do not decide to call a schema validator, we can skip adapting these (and save some time) and even remove them. Let's discuss with @spyrostz .

gsy-contracts/contracts/MarketController.sol
spyrostz: Agreed and created ticket https://gridsingularity.atlassian.net/browse/DD-439

gsy-market-orchestrator/src/community_source.rs
spyrostz: The main problem that I see here is that the trait is used only once, in the CommunityProvider implementation. To my knowledge, traits are generally used as "interfaces" or "abstract base classes", so essentially collects function definitions that force the structs that implement the trait to follow these function definitions. This makes sense if we have multiple classes that all need to export a similar functionality with different implementation. 
Here though, there is a single class that handles both EWDS and HTTP. If we had different classes for each, then the trait would be the ideal way to force a common contract across both classes, in this case though it is really not needed, since the one class can define its "contract" via its public methods. 
To summarise, I think that the trait should be deleted, unless there is a strong suspicion that more classes will be needed in the future

hannesdiedrich: We need a universal structure that calls the offchain storage endpoints (depending on the http and ewds setting). So if re-written. this trait could be re-used several times (I already have two PRs open where I could re-use) We somehow have to abstract the endpoint urls and the de-serialsation away and then this would work.

gsy-market-orchestrator/src/orchestrator.rs
```diff
@@ -1,15 +1,19 @@
...
+use primitives::{
+    constants::GLOBAL_CONSTANTS,
+    utils::{generate_market_id, timestamp_to_datetime_string},
+};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tokio::time::sleep;
 use tracing::{error, info, warn};
 
-pub async fn run<C>(config: Config, client: C) -> anyhow::Result<()>
+pub async fn run<C, S>(config: Config, client: C, community_source: S) -> anyhow::Result<()>
```
spyrostz: If I read the code correctly, we would not need the second generic argument here if we remove the trait.

gsy-offchain-storage/src/ewds_handler.rs
```diff
@@ -148,20 +150,30 @@ pub async fn start_ewds_request_handler(db: DatabaseWrapper, config: EwdsHandler
         config.gateway_url, config.request_fqcn, config.response_fqcn
     );
 
-    let client = Client::new();
+    let workers = EwdsOperation::ALL
+        .into_iter()
+        .map(|operation| run_topic_worker(db.clone(), Client::new(), config.clone(), operation));
+    join_all(workers).await;
```
spyrostz: Nothing to do here, but I am concerned that listening to one topic per worker would incur significant traffic to the EWDS, and contribute to their performance issues. If needed in the future, we can see whether they support wildcard topics, so that we listen to all topics with one / a few workers.

gsy-offchain-storage/src/ewds_handler.rs
spyrostz: https://gridsingularity.atlassian.net/browse/DD-439

gsy-market-orchestrator/src/community_source.rs
keyanhov: I agree with Hannes about sharing the HTTP/EWDS client across services. That client could encapsulate URLs, operations, and deserialization, while the orchestrator keeps its small CommunityProvider interface for testing. I suggest coordinating that extraction with another ticket and PRs.

gsy-market-orchestrator/src/orchestrator.rs
keyanhov: Correct. Accepting the concrete source would remove S. Currently, S: CommunityProvider allows the real source and the test mock to be passed through to the orchestration logic

gsy-offchain-storage/src/ewds_handler.rs
keyanhov: That's true with eight operations and the default 500 ms interval, idle polling can approach 16 requests/second per handler instance, excluding response traffic. Separate workers prevent one topic’s processing/backoff from delaying the others, but the polling overhead scales with topic count. We should check whether the gateway supports polling multiple topics in one request increasing the interval or backing off on empty polls are also options

schemas/ewds/intelligent/int.communities.query.request.v1.json
spyrostz: Created a task for this to discuss in our sprint planning https://gridsingularity.atlassian.net/browse/DD-441

spyrostz (approved):
LGTM, opened tickets for followup items.

gsy-market-orchestrator/src/community_source.rs
spyrostz: I also agree, however I do not see the trait being used even then assuming that we maintain this struct. For the trait to be used, I would expect to have 2 structs, one for HTTP (e.g. OffchainStorageCommunityHTTP) and one for EWDS (e.g. OffchainStorageCommunityEWDS), each implementing the trait. That way, the implementation of EWDS and HTTP would be isolated in their own struct (assuming that the trait would be augmented with methods such as fetch(), with different implementation for HTTP and EWDS), which would justify the use of the trait. In this implementation though, you use separate methods fetch_via_http and fetch_via_ewds, therefore the trait is not really used. 
TLDR, I am fine as is, but especially when this struct is moved to primitives it would be great if either the trait was removed, or to be actually used by creating separate structs (and I would very much favor the latter).

keyanhov: Agreed. Just to clarify, the current trait is used by MockCommunityProvider in the orchestrator tests, allowing us to test fetch failures and empty results without network services. We can keep this PR as is and revisit the transport structure in the shared-client ticket

spyrostz: Agreed as well and makes sense to refactor it in the followup ticket and not here.

spyrostz:
> I left some comments, will review again after answered/adapted.
> 
> I have a general comment, mainly for @spyrostz: We need to discuss/design and agree on the process which is creating entries in the `MarketSchema` document in the DB. Or am I missing that this is already done?

Hi @hannesdiedrich this needs to be implemented in EVM Handler in offchain storage, IIRC our agreement around a month ago. We will revisit when adapting it.

